PR #1, opened 2019-10-19
EPAMLABSBRN-65

ElenaSpb:
EPAMLABSBRN-65
add ExceptionControllerAdvice


src/main/kotlin/com/epam/brn/dto/ErrorResponse.kt
```diff
@@ -0,0 +1,3 @@
+package com.epam.brn.dto
+
+data class ErrorResponse(val message: String?)
```
goopnigoop: I've got a suggestion to add here field code, to keep here error code

PR #3, opened 2019-10-21
blazeUI usage

iki6:
[EPMLABSBRN-60](https://jira.epam.com/jira/browse/EPMLABSBRN-60)

**Utilized** blaze-ui components for the app
image

### DIFF
--- a/frontend/app/components/exercise-navigation/template.hbs
+++ b/frontend/app/components/exercise-navigation/template.hbs
@@ -1,12 +1,16 @@
-{{#each (sort-by "order" @tasks) as |task index|}}
-  {{#if (eq 0 index)}}
-    Задачи:
-  {{/if}}
+<nav class="c-pagination">
+  <div class="c-pagination__pages tasks-pages-pagination">
+    {{#each (sort-by "order" @tasks) as |task index|}}
+      {{#if (eq 0 index)}}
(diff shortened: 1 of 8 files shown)


frontend/app/components/exercise-navigation/template.hbs
```diff
@@ -1,12 +1,16 @@
...
-{{/each}}
+      {{#link-to "series.exercise.task" task putActiveAttr=true class="c-pagination__control"}}
+        {{~task.order~}}
+      {{/link-to}}
+      &nbsp;
+    {{else}}
+      Пока в этом упражнении нет задач 🤷‍♂️
+    {{/each}}
+  </div>
+</nav>
```
H1D: @iki6 can't we use `<blaze-pagination>` tag?
"JavaScript" tab here – https://www.blazeui.com/components/pagination

ElenaSpb:
could you please name MR like

EPMLABSBRN-TaskNumber
Description: what was done

frontend/app/components/exercise-navigation/template.hbs
iki6: @H1D We can, but it'll look like this:

image

and according on data in the docs all the attributes we can pass are the total count of pages and the current one( without number of shown pages or showArrows flag )

PR #11, opened 2019-10-22
EPMLABSBRN-63 create Unique KEY StudyHistory (userId and exerciseId) and update record

rpcep:
**EPMLABSBRN-63**  Create Unique KEY StudyHistory (userId and exerciseId) and update record

**CONTENTS:**
- Configured mapstruct processor, KAPT for code generation and idea plugin. Necessary for MapStruct to work.
- Added PATCH and PUT. Added PUT capabilitites to POST when record exists as requested.
- Created constraints.


rpcep:
Ready for review.

src/main/kotlin/com/epam/brn/dto/StudyHistoryDto.kt
```diff
@@ -1,21 +1,18 @@
 package com.epam.brn.dto
 
 import java.time.LocalDateTime
-import javax.validation.constraints.NotNull
 
 class StudyHistoryDto(
```
goopnigoop: Maybe make it data class too?

src/main/kotlin/com/epam/brn/dto/StudyHistoryDto.kt
```diff
@@ -1,21 +1,18 @@
...
-    val repetitionCount: Short = 0
+    var userId: Long?,
+    var exerciseId: Long?,
+    var startTime: LocalDateTime?,
+    var endTime: LocalDateTime?,
+    var doneTasksCount: Short?,
+    var successTasksCount: Short?,
+    var repetitionCount: Short?
 ) {
+    constructor() : this(null, null, null, null, null, null, null)
```
goopnigoop: Can you explain why do we need that constructor?

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -15,12 +17,18 @@ class StudyHistoryService(
...
     @Autowired val entityManager: EntityManager
 ) {
-    private val log = logger()
-
-    fun saveStudyHistory(studyHistoryDto: StudyHistoryDto): Long? {
+    fun saveOrReplaceStudyHistory(studyHistoryDto: StudyHistoryDto): Long? {
+        val studyHistoryEntityOptional = studyHistoryRepository.findByUserAccount_IdAndExercise_Id(
+            studyHistoryDto.userId,
+            studyHistoryDto.exerciseId
+        )
```
goopnigoop: what do you think to simplify this?
 val studyHistoryEntityOptional = studyHistoryRepository.findByUserAccount_IdAndExercise_Id(
            studyHistoryDto.userId,
            studyHistoryDto.exerciseId
        ).ifPresent(entity -> updateEntity(studyHistoryDto,entity))

src/main/kotlin/com/epam/brn/dto/StudyHistoryDto.kt
rpcep: MapStruct needs it. From official example for kotlin:
https://github.com/mapstruct/mapstruct-examples/blob/master/mapstruct-kotlin/src/main/kotlin/org/mapstruct/example/kotlin/model/Person.kt

src/main/kotlin/com/epam/brn/dto/StudyHistoryDto.kt
rpcep: Could you elaborate on that?

goopnigoop: I mean our dto classes are kotlin data [classes](https://kotlinlang.org/docs/reference/data-classes.html) 
Or we don't need Dto to be data classes? Advantages of data classes are by the link

rpcep: Got it, added the changes.

rpcep:
Fixed merge conflictsd

src/test/kotlin/com/epam/brn/service/StudyHistoryServiceTest.kt
```diff
@@ -44,20 +43,44 @@ internal class StudyHistoryServiceTest {
             endTime = LocalDateTime.now(),
             exerciseId = 1L
         )
+        val entity = StudyHistory(
+            id = null,
```
ElenaSpb: why it is need to set id here?

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -15,12 +17,18 @@ class StudyHistoryService(
     @Autowired val studyHistoryRepository: StudyHistoryRepository,
     @Autowired val entityManager: EntityManager
 ) {
-    private val log = logger()
```
ElenaSpb: could you please add some useful logging for each function 
log.info("...")
it could help us later

src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
```diff
@@ -3,9 +3,10 @@ package com.epam.brn.repo
 import com.epam.brn.model.StudyHistory
 import org.springframework.data.repository.CrudRepository
 import org.springframework.stereotype.Repository
+import java.util.Optional
 
 @Repository
 interface StudyHistoryRepository : CrudRepository<StudyHistory, Long> {
 
-    fun findByIdLike(id: String): List<StudyHistory>
+    fun findByUserAccount_IdAndExercise_Id(userId: Long?, exerciseId: Long?): Optional<StudyHistory>
```
ElenaSpb: could it be without `_` ?

rpcep: No, it cannot, it is hibernate syntax.

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
```diff
@@ -22,16 +29,18 @@ data class StudyHistory(
     val id: Long? = null,
     @OneToOne(cascade = [(CascadeType.ALL)])
     @JoinColumn(name = "user_id")
-    val userAccount: UserAccount,
+    var userAccount: UserAccount?,
```
ElenaSpb: just does it mean that userAccount is not required?

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
rpcep: At most it should be log.debug, but i do not think it is a good idea to clutter the application with logs in services. I propose we do debug log in controllers, what do you think?

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
```diff
@@ -22,16 +29,18 @@ data class StudyHistory(
...
     @OneToOne(cascade = [(CascadeType.ALL)])
     @JoinColumn(name = "exercise_id")
-    val exercise: Exercise? = null,
-    val startTime: LocalDateTime,
-    val endTime: LocalDateTime,
-    val doneTasksCount: Short,
-    val successTasksCount: Short,
-    val repetitionCount: Short
+    var exercise: Exercise? = null,
```
ElenaSpb: just does it mean that exercise is not required?

src/main/kotlin/com/epam/brn/dto/StudyHistoryDto.kt
```diff
@@ -1,21 +1,18 @@
...
-    @NotNull
-    val startTime: LocalDateTime,
-    @NotNull
-    val endTime: LocalDateTime,
-    val doneTasksCount: Short = 0,
-    val successTasksCount: Short = 0,
-    val repetitionCount: Short = 0
+data class StudyHistoryDto(
+    var userId: Long?,
+    var exerciseId: Long?,
```
ElenaSpb: why this two needable parameters are not required

rpcep: If we want to use MapStruct, we need nullable fields. I think that the tradeoff is worth it, something has to give.
Else we need to either map manually or find another solution.

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
rpcep: If we want to use MapStruct, we need nullable fields. I think that the tradeoff is worth it, something has to give.
Else we need to either map manually or find another solution.

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
rpcep: If we want to use MapStruct, we need nullable fields. I think that the tradeoff is worth it, something has to give.
Else we need to either map manually or find another solution.

ElenaSpb:
great work!

rpcep:
Thanks!

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
ElenaSpb: may be there are some annotations which can make semantically restriction that this fields are required - something lwith nullable = false?

goopnigoop:
> MapStruct creates some constraints in data objects, but i think the benefits in development speed and ease are worth it. Else we need to either map manually or find another solution.
> @goopnigoop Any thoughts on this?

I've used this one in java project and it's more preferable than boilerplate code. I like to see it here, I think if we faced some problems or restrictions we can rollback to simple kotlin solution

src/main/kotlin/com/epam/brn/converter/StudyHistoryConverter.kt
```diff
@@ -0,0 +1,13 @@
+package com.epam.brn.converter
+
+import com.epam.brn.dto.StudyHistoryDto
+import com.epam.brn.model.StudyHistory
+import org.mapstruct.Mapper
+import org.mapstruct.MappingTarget
+import org.mapstruct.NullValuePropertyMappingStrategy
+
+@Mapper(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_NULL)
```
goopnigoop: I think we have to use componentModel property here to enable Dependency Injection

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -30,8 +38,40 @@ class StudyHistoryService(
...
+            studyHistoryDto.exerciseId
+        ).get()
+        return updateEntityNotNullOnly(studyHistoryDto, studyHistoryEntity)
+    }
+
+    private fun updateEntityNotNullOnly(
+        studyHistoryDto: StudyHistoryDto,
+        studyHistoryEntity: StudyHistory
+    ): Long? {
+        Mappers.getMapper(StudyHistoryNotNullConverter::class.java)
```
goopnigoop: I think we have to use spring's DI and autowire this mappers
https://mapstruct.org/documentation/stable/reference/html/#using-dependency-injection

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
rpcep: Done

goopnigoop: @OneToOne there is property optional, could be false

rpcep:
Finished requested changes, please review when possible

src/main/kotlin/com/epam/brn/converter/StudyHistoryConverter.kt
rpcep: Done

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
rpcep: Done

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -30,8 +43,42 @@ class StudyHistoryService(
...
+        ).get()
+        return updateEntity(studyHistoryDto, studyHistoryEntity)
+    }
+
+    fun patchStudyHistory(studyHistoryDto: StudyHistoryDto): Long? {
+        log.debug("Patching $studyHistoryDto")
+        val studyHistoryEntity = studyHistoryRepository.findByUserAccount_IdAndExercise_Id(
+            studyHistoryDto.userId,
+            studyHistoryDto.exerciseId
+        ).get()
```
goopnigoop: What do you think about avoiding use get() in optional, and use .map ({ updateEntityNotNullOnly(studyHistoryDto, it) }) .orElseThrow( blah blah blah)?
Because in case optional is empty get() throws an exception anyway, but this way flow fill be visible

src/main/kotlin/com/epam/brn/converter/StudyHistoryNotNullConverter.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.epam.brn.dto.StudyHistoryDto
+import com.epam.brn.model.StudyHistory
+import org.mapstruct.InjectionStrategy
+import org.mapstruct.Mapper
+import org.mapstruct.MappingTarget
+import org.mapstruct.NullValuePropertyMappingStrategy
+
+@Mapper(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE, componentModel = "spring", injectionStrategy = InjectionStrategy.CONSTRUCTOR)
+interface StudyHistoryNotNullConverter {
```
goopnigoop: I'm not sure but what if use only one converter instead of 2 with different methods and on methods set @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE,
            nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS)
something like this, took this annotation from google

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
rpcep: Good idea, done

src/main/kotlin/com/epam/brn/converter/StudyHistoryNotNullConverter.kt
rpcep: Done

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -13,25 +14,68 @@ import javax.persistence.EntityManager
...
+            ).id
+        }
+    }
+
+    fun replaceStudyHistory(studyHistoryDto: StudyHistoryDto): Long? {
+        log.debug("Replacing $studyHistoryDto")
+        val studyHistoryEntity = studyHistoryRepository.findByUserAccount_IdAndExercise_Id(
+            studyHistoryDto.userId,
+            studyHistoryDto.exerciseId
+        ).get()
```
goopnigoop: I mean this get() and below:

What do you think about avoiding use get() in optional, and use .map ({ updateEntityNotNullOnly(studyHistoryDto, it) }) .orElseThrow( blah blah blah)?
Because in case optional is empty get() throws an exception anyway, but this way flow fill be visible

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
```diff
@@ -9,8 +9,15 @@ import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.OneToOne
 import javax.persistence.SequenceGenerator
+import javax.persistence.Table
+import javax.persistence.UniqueConstraint
 
 @Entity
+@Table(
+    uniqueConstraints = [
+        UniqueConstraint(columnNames = ["user_id", "exercise_id"])
```
RiveriRiv: What do you think about adding an index for this values?

rpcep: Done

src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
rpcep: Done

PR #12, opened 2019-10-22
    EPMLABSBRN-79  make prod profile to start it on remote instance and write command to start it in jenkins

rpcep:
**EPMLABSBRN-79** Make prod profile to start it on remote instance and write command to start it in jenkins

**CONTENTS**
- Changed environment variable name "PG_*" to "POSTGRES_*" to have the same variable name for the same content across containers.
- Added new prod profile.
- Added profile activation from environment.
- Further structured config, default configuration is applied to all profiles, then it is overriden by specific profile configuration.
- Added default passwords in dev and local profiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ To install docker:
 
 To run docker use the following command:
 *  on linux:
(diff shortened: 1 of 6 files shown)


rpcep:
Ready for review

PR #17, opened 2019-10-23
[EPMLABSBRN-77] - Added uniqe constraint and index for word and audioFileUrl resource attributes

RiveriRiv:


src/main/kotlin/com/epam/brn/model/Resource.kt
```diff
@@ -2,13 +2,20 @@ package com.epam.brn.model
...
-import javax.persistence.ManyToMany
+import javax.persistence.GeneratedValue
 import javax.persistence.Column
+import javax.persistence.ManyToMany
+import javax.persistence.GenerationType
 
 @Entity
+@Table(
+    uniqueConstraints = [UniqueConstraint(columnNames = ["word", "audioFileUrl"])],
+    indexes = [Index(name = "word_audio_file_idx", columnList = "word, audioFileUrl")]
```
goopnigoop: suggestion lets add one more index only for audioFileUrl too

src/main/kotlin/com/epam/brn/model/Resource.kt
```diff
@@ -17,9 +24,7 @@ data class Resource(
     val audioFileUrl: String,
```
ElenaSpb: as audioFileUrl is requered - maybe add this annotation
@Column(nullable = false)
to its too?

src/main/kotlin/com/epam/brn/model/Resource.kt
ElenaSpb: maybe for index for word only - as we will find recoueses for words

RiveriRiv: done

PR #20, opened 2019-10-24
EPMLABSBRN-62 Get end-point: get available exercises for user

rpcep:
**EPMLABSBRN-62** Get end-point: get available exercises for user

**Description:**
- Implemented endpoint.
- Covered with tests

### DIFF
--- a/src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
+++ b/src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
@@ -1,23 +1,33 @@
 package com.epam.brn.controller
 
+import com.epam.brn.constant.BrnPath
 import com.epam.brn.dto.ExerciseDto
 import com.epam.brn.service.ExerciseService
-import com.epam.brn.constant.BrnPath
+import io.swagger.annotations.Api
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
 @RequestMapping(BrnPath.EXERCISE)
+@Api(value = BrnPath.EXERCISE, description = "Contains the results of a finished exercise")
 class ExerciseController(@Autowired val exerciseService: ExerciseService) {
 
     @GetMapping
-    fun getExercises(
-        @RequestParam(value = "name", defaultValue = "0") name: String,
-        @RequestParam(value = "userId", defaultValue = "0") userId: String
+    fun getExercisesByName(
+        @RequestParam(value = "name", required = true) name: String
     ): List<ExerciseDto> {
         return exerciseService.findExercises(name)
+    }
+
+    @RequestMapping(value = [BrnPath.USER + "/{userID}"], method = [RequestMethod.GET])
+    fun getAvailableExercises(
+        @PathVariable("userID") userID: Long
+    ): List<ExerciseDto> {
+        return exerciseService.findAvailableExercises(userID)
     }
 }
--- a/src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
+++ b/src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
@@ -7,5 +7,5 @@ import org.springframework.stereotype.Repository
 @Repository
 interface StudyHistoryRepository : CrudRepository<StudyHistory, Long> {
 
-    fun findByIdLike(id: String): List<StudyHistory>
+    fun findByUserAccount_Id(id: Long?): List<StudyHistory>
 }
--- a/src/main/kotlin/com/epam/brn/service/ExerciseService.kt
+++ b/src/main/kotlin/com/epam/brn/service/ExerciseService.kt
@@ -2,17 +2,27 @@ package com.epam.brn.service
 
 import com.epam.brn.dto.ExerciseDto
 import com.epam.brn.repo.ExerciseRepository
+import com.epam.brn.repo.StudyHistoryRepository
 import org.apache.logging.log4j.kotlin.logger
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
+import java.util.Optional
 
 @Service
-class ExerciseService(@Autowired val exerciseRepository: ExerciseRepository) {
-
+class ExerciseService(
+    @Autowired val exerciseRepository: ExerciseRepository,
+    @Autowired val studyHistoryRepository: StudyHistoryRepository
+) {
     private val log = logger()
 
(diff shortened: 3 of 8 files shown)


src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -1,23 +1,33 @@
...
-        @RequestParam(value = "name", defaultValue = "0") name: String,
-        @RequestParam(value = "userId", defaultValue = "0") userId: String
+    fun getExercisesByName(
+        @RequestParam(value = "name", required = true) name: String
     ): List<ExerciseDto> {
         return exerciseService.findExercises(name)
     }
+
+    @RequestMapping(value = [BrnPath.USER + "/{userID}"], method = [RequestMethod.GET])
+    fun getAvailableExercises(
```
goopnigoop: is it proper behavior to get exercises by path "/user"? Maybe get exercises for user using path "/exercises"?

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -2,17 +2,27 @@ package com.epam.brn.service
...
 
     fun findExercises(name: String): List<ExerciseDto> {
         val exercises = exerciseRepository.findByNameLike(name)
         return exercises.map { exercise -> exercise.toDto() }
     }
+
+    fun findAvailableExercises(userID: Long): List<ExerciseDto> {
+        log.debug("Searching available exercises for $userID")
+        val history = studyHistoryRepository.findByUserAccount_Id(userID)
+        return history.map { h -> Optional.ofNullable(h.exercise).map { e -> e.toDto() }.orElse(null) }
```
goopnigoop: I've gon a suggestion:

lets add to our project apache commons collections:
    compile group: "org.apache.commons", name: "commons-collections4", version: "4.0"
There are a bunch of great methods
and simplify this line like:

   return emptyIfNull(history)
            .mapNotNull { it.exercise }
            .map { it.toDto() }

and this way we will return empty list instead of null

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
rpcep: It is adding to the class mapper, as in "/exercise/user/{userID}".

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
rpcep: Agree. Done.

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
goopnigoop: sorry, i missed this one: RequestMapping(BrnPath.EXERCISE)

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -2,17 +2,27 @@ package com.epam.brn.service
...
 
     fun findExercises(name: String): List<ExerciseDto> {
         val exercises = exerciseRepository.findByNameLike(name)
         return exercises.map { exercise -> exercise.toDto() }
     }
+
+    fun findAvailableExercises(userID: Long): List<ExerciseDto> {
+        log.debug("Searching available exercises for $userID")
+        val history = studyHistoryRepository.findByUserAccount_Id(userID)
+        return emptyIfNull(history).mapNotNull { it.exercise }.map { it.toDto() }
```
goopnigoop: i think the second map should be mapNotNull too missed it

rpcep: Changed.

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -2,17 +2,27 @@ package com.epam.brn.service
...
     private val log = logger()
 
     fun findExercises(name: String): List<ExerciseDto> {
         val exercises = exerciseRepository.findByNameLike(name)
         return exercises.map { exercise -> exercise.toDto() }
     }
+
+    fun findAvailableExercises(userID: Long): List<ExerciseDto> {
+        log.debug("Searching available exercises for $userID")
+        val history = studyHistoryRepository.findByUserAccount_Id(userID)
```
ElenaSpb: please lets name functions without `_`

rpcep: That is hibernate syntax. It cannot be changed. The only other way is using explicit queries.

rpcep: https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-property-expressions

That syntax is correct.

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -2,17 +2,27 @@ package com.epam.brn.service
...
+    @Autowired val studyHistoryRepository: StudyHistoryRepository
+) {
     private val log = logger()
 
     fun findExercises(name: String): List<ExerciseDto> {
         val exercises = exerciseRepository.findByNameLike(name)
         return exercises.map { exercise -> exercise.toDto() }
     }
+
+    fun findAvailableExercises(userID: Long): List<ExerciseDto> {
```
ElenaSpb: semattically there we have findDoneExercises as I see, 
so it would be good to name it as it do

rpcep: Changed

src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
```diff
@@ -8,5 +8,6 @@ import java.util.Optional
 @Repository
 interface StudyHistoryRepository : CrudRepository<StudyHistory, Long> {
 
+    fun findByUserAccount_Id(id: Long?): List<StudyHistory>
```
ElenaSpb: we really do not need to use `_` 
as I told and then I checked - it works without this `_` too
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-creation

so lets remove all `_` from functions names

rpcep: That is not correct. From documentation:


> List<Person> findByAddressZipCode(ZipCode zipCode);
> Although this should work for most cases, it is possible for the algorithm to select the wrong property. Suppose the Person class has an addressZip property as well. The algorithm would match in the first split round already, choose the wrong property, and fail (as the type of addressZip probably has no code property).
> To resolve this ambiguity you can use _ inside your method name to manually define traversal points. 
> List<Person> findByAddress_ZipCode(ZipCode zipCode);
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-property-expressions

I do not agree with removing the underscore.

rpcep: Changed.

PR #26, opened 2019-10-26
EPAMLABSBRN-94 remake series end-point to support new naming conversion

ElenaSpb:
EPAMLABSBRN-94 remake series end-point to support new naming conversion

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/SeriesService.kt
+++ b/src/main/kotlin/com/epam/brn/service/SeriesService.kt
@@ -11,16 +11,19 @@ class SeriesService(private val seriesRepository: SeriesRepository) {
 
     private val log = logger()
 
-    fun findSeriesForGroup(groupId: Long): List<SeriesDto> {
+    fun findSeriesForGroup(groupId: Long, include: String): List<SeriesDto> {
         log.debug("try to find series for groupId=$groupId")
         val series = seriesRepository.findByExerciseGroupLike(groupId)
-        return series.map { seriesEntry -> seriesEntry.toDto() }
+        if ("exercises" == include)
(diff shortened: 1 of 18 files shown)


src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -11,8 +12,15 @@ class SeriesService(private val seriesRepository: SeriesRepository) {
...
+        log.debug("try to find series for groupId=$groupId")
         val series = seriesRepository.findByExerciseGroupLike(groupId)
-        log.info("try to find series for groupId=$groupId")
         return series.map { seriesEntry -> seriesEntry.toDto() }
     }
+
+    fun findSeriesForId(seriesId: Long): SeriesDto {
+        log.debug("try to find series for seriesId=$seriesId")
+        val series = seriesRepository.findById(seriesId)
+            .orElseGet { throw NoDataFoundException("no series was found for id=$seriesId") }
```
rpcep: When the series is not found, JPA throws an excaption. You are covering the original exception with this one, i would either include the original exception in this one or just let it surface and be handled by advices.

ElenaSpb: there may be no any original exception, just no series for this id

rpcep: True, but if there is we will never know which one was.

ElenaSpb: this variant works good

ElenaSpb: Ruben, I have fixed it, yes

PR #29, opened 2019-10-29
Epmlabsbrn 96 - Backend: remake end-point get available exercises for user as described in contract

rpcep:
[EPMLABSBRN-96](https://jira.epam.com/jira/browse/EPMLABSBRN-96) Backend: remake end-point get available exercises for user as described in contract

**Description**:
- Modified endpoint to conform contract.
- removed unused get by name.
- Created new endpoint get by id.
- Added integration tests.

### DIFF
--- a/src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
+++ b/src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
@@ -1,16 +1,15 @@
 package com.epam.brn.controller
 
 import com.epam.brn.constant.BrnPath
-import com.epam.brn.dto.ExerciseDto
+import com.epam.brn.dto.ExerciseDtoWrapper
 import com.epam.brn.service.ExerciseService
 import io.swagger.annotations.Api
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation.GetMapping
-import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
+import java.util.*
 
 @RestController
 @RequestMapping(BrnPath.EXERCISES)
@@ -20,14 +19,21 @@ class ExerciseController(@Autowired val exerciseService: ExerciseService) {
     @GetMapping
     fun getExercisesByName(
         @RequestParam(value = "name", required = true) name: String
-    ): List<ExerciseDto> {
-        return exerciseService.findExercises(name)
+    ): ExerciseDtoWrapper {
+        return ExerciseDtoWrapper(data = exerciseService.findExercisesByNameLike(name))
     }
 
-    @RequestMapping(value = ["/user/{userID}"], method = [RequestMethod.GET])
+    @GetMapping
+    fun getExercisesByID(
+        @RequestParam(value = "userID", required = true) userID: Long
+    ): ExerciseDtoWrapper {
+        return ExerciseDtoWrapper(data = Collections.singletonList(exerciseService.findExercisesByID(userID)))
+    }
+
+    @GetMapping
     fun getDoneExercises(
-        @PathVariable("userID") userID: Long
-    ): List<ExerciseDto> {
-        return exerciseService.findDoneExercises(userID)
+        @RequestParam(value = "userID", required = true) userID: Long
+    ): ExerciseDtoWrapper {
+        return ExerciseDtoWrapper(data = exerciseService.findDoneExercises(userID))
     }
 }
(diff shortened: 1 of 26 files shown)


src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -1,33 +1,46 @@
...
+        @RequestParam(value = "exerciseID", required = false) exerciseID: Long?,
+        @RequestParam(value = "userID", required = false) userID: Long?
+    ): ExerciseDtoWrapper {
+        if (ObjectUtils.isNotEmpty(exerciseID).xor(ObjectUtils.isNotEmpty(userID))) {
+            if (ObjectUtils.isNotEmpty(exerciseID)) {
+                return getExercisesByID(exerciseID!!)
+            }
+            return getDoneExercises(userID!!)
+        }
+        throw InvalidParametersException("Only exerciseID or userID is allowed, but one is required")
```
RiveriRiv: I think it would be better to avoid checking exerciseID several times, for example:
if (ObjectUtils.isNotEmpty(exerciseID) {
    return getExercisesByID(exerciseID!!)
} 
if (ObjectUtils.isNotEmpty(userID)) {
    return getDoneExercises(userID!!)
}

throw InvalidParametersException("Only exerciseID or userID is allowed, but one is required")

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -1,33 +1,46 @@
...
+    }
+
+    fun getExercisesByID(
```
RiveriRiv: Could it be private?

ElenaSpb: it would be difficult to write tests on private methods

rpcep: Removed

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
rpcep: Does not apply, removed.

src/main/kotlin/com/epam/brn/dto/BaseResponseWrapper.kt
```diff
@@ -0,0 +1,6 @@
+package com.epam.brn.dto
+
+open class BaseResponseWrapper {
+    var error: ArrayList<String> = ArrayList()
```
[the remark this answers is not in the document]
rpcep: Does not apply, removed.

src/main/kotlin/com/epam/brn/dto/BaseResponseWrapper.kt
```diff
@@ -0,0 +1,6 @@
...
+    var metadata: ArrayList<String> = ArrayList()
```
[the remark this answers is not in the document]
rpcep: Does not apply, removed.

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
rpcep: Does not apply, removed.

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -15,14 +16,15 @@ class ExerciseService(
 ) {
     private val log = logger()
 
-    fun findExercises(name: String): List<ExerciseDto> {
-        val exercises = exerciseRepository.findByNameLike(name)
-        return exercises.map { exercise -> exercise.toDtoWithTasks() }
+    fun findExercisesByID(exerciseID: Long): ExerciseDto {
```
goopnigoop: I'd suggest to rename it to findExerciseByID, not the plural

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -11,23 +13,24 @@ import org.springframework.web.bind.annotation.RequestMapping
...
-    @RequestMapping(value = ["/user/{userID}"], method = [RequestMethod.GET])
-    fun getDoneExercises(
-        @PathVariable("userID") userID: Long
-    ): List<ExerciseDto> {
-        return exerciseService.findDoneExercises(userID)
+    @RequestMapping(value = ["/{$EXERCISE_ID}"], method = [RequestMethod.GET])
+    fun getExercisesByID(
+        @PathVariable(EXERCISE_ID) exerciseID: Long
+    ): BaseResponseDto {
+        return BaseResponseDto(data = Collections.singletonList(exerciseService.findExercisesByID(exerciseID)))
```
goopnigoop: what if use kotlin's listOf() that wraps it in an immutable list, or do you really need a singleton list?

rpcep: I checked the contract and, in fact, it should not be a list at all. 
Also, old habits die hard :). Will use listOf() from now on.
Changed.

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
rpcep: Changed

PR #32, opened 2019-11-02
EPMLABSBRN-105  Back-end: support child ids by defualt in GET requests for groups&series

ElenaSpb:
EPMLABSBRN-105 Back-end: support child ids by defualt in GET requests for groups&series

Description
response include child ids

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/SeriesService.kt
+++ b/src/main/kotlin/com/epam/brn/service/SeriesService.kt
@@ -1,6 +1,5 @@
 package com.epam.brn.service
 
-import com.epam.brn.dto.SeriesDto
 import com.epam.brn.exception.NoDataFoundException
 import com.epam.brn.repo.SeriesRepository
 import org.apache.logging.log4j.kotlin.logger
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -12,24 +11,22 @@ class SeriesService(private val seriesRepository: SeriesRepository) {
...
-        return listDto
+            return series.map { seriesEntry -> seriesEntry.toDtoWithExerciseIds() }
     }
 
-    fun findSeriesForId(seriesId: Long, include: String): SeriesDto {
+    fun findSeriesForId(seriesId: Long, include: String): Any {
         log.debug("try to find series for seriesId=$seriesId")
         val series = seriesRepository.findById(seriesId)
             .orElseThrow { NoDataFoundException("no series was found for id=$seriesId") }
         if (include == EXERCISES)
```
goopnigoop: as we discussed with @H1D we decided to proceed implementation without included flag

ElenaSpb: this  is already was done after first suggestion with `include` flag,
so as we can return to it in future - lets is stay, 
ispessially it works correct

PR #35, opened 2019-11-03
EPMLABSBRN-37 upgrade dockerization

Mikhail-Sosnin:
[EPMLABSBRN-37](https://jira.epam.com/jira/browse/EPMLABSBRN-37)

**Description**:
Application Dockerization tuning.
Extracting unit tests execution to the separate compose file leads to the fact that we have a single point of truth in all possible situations when we have different results in test execution (at the environment side vs on local machines under Windows/Linux/macOS).
Makefile should make our workflow a bit more comfortable, especially in the case when someone a bit non-familiar with some specific commands and keys and we need to save our time and just use 
**make [command]** instead of a bit longer one.

### DIFF
--- a/src/main/resources/application-dev.properties
+++ b/src/main/resources/application-dev.properties
@@ -1,6 +1,6 @@
 spring.datasource.username=${POSTGRES_USER:admin}
 spring.datasource.password=${POSTGRES_PASSWORD:admin}
-spring.datasource.url=jdbc:postgresql://localhost:5432/${POSTGRES_DB:brn}
+spring.datasource.url=jdbc:postgresql://${DB_URL:db_brn-test}:5432/${POSTGRES_DB:brn}
 spring.jpa.hibernate.ddl-auto=create
 
 logging.level.org.springframework=debug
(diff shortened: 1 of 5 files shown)


src/main/resources/application-dev.properties
```diff
@@ -1,6 +1,6 @@
 spring.datasource.username=${POSTGRES_USER:admin}
 spring.datasource.password=${POSTGRES_PASSWORD:admin}
-spring.datasource.url=jdbc:postgresql://localhost:5432/${POSTGRES_DB:brn}
+spring.datasource.url=jdbc:postgresql://${DB_URL:db_brn-test}:5432/${POSTGRES_DB:brn}
```
ElenaSpb: db from localhost is convinient to use on local envarment, that is why it was set.
why you want to change it.. ?

Mikhail-Sosnin: @ElenaSpb you are right. Reverted localhost and extracted DB_URL to the compose env vars.

ElenaSpb: we use dev profile only on our developers machine,
and localhost is the way how we run application (and how it is described in our current Readme file),
so we need not change it,
or lets set default value of this prod in application.properties

rpcep: This is the new value, in last push:
spring.datasource.url=jdbc:postgresql://${DB_URL:localhost}:5432/${POSTGRES_DB:brn} 

${DB_URL:localhost} means it will use the environment variable if present or localhost else. The environment variable is set in docker compose, so it will work well in both docker compose mode and in intellij debug mode. 

I think it is the correct config.

ElenaSpb: yes, I see

PR #36, opened 2019-11-05
EPMLABSBRN-101 Add availability on exercises retrieved on GET /exercises

rpcep:
[EPMLABSBRN-101](https://jira.epam.com/jira/browse/EPMLABSBRN-101) Add availability on exercises retrieved on GET /exercises

**Description**:
- Now exercises endpoint brings all exercises available to user or user/series pair.
- Each exercise has "available" property that indicates if the exercise is done.
- Moved initialization of tables for resources loading to dev initialization runner.
- Batch runner will not run in dev mode.


src/main/kotlin/com/epam/brn/config/DevConfig.kt
```diff
@@ -92,4 +106,87 @@ class DevConfig(
         )
         log.debug("---- Created $studyHistory")
     }
+
+    fun loadInitialDBDataForFileUploadDev() {
```
ElenaSpb: could you please see LoadHandBook for load initial data nowdays,
future data would be prepared in file and load from it on start application,
so load data in config I think is not very good idea

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -22,10 +23,15 @@ class ExerciseController(@Autowired val exerciseService: ExerciseService) {
 
     @GetMapping
     @ApiOperation("Get done exercises for user")
-    fun getExercisesByUserID(
-        @RequestParam(value = USER_ID) userId: Long
+    fun getExercises(
+        @RequestParam(value = USER_ID, required = true) userId: Long,
+        @RequestParam(value = SERIES_ID, required = false) seriesId: Long?
```
goopnigoop: I'm not sure, but when I was implementing the task controller @ElenaSpb made me set exerciseId as required and avoid getting tasks without exercise id. There were a lot of discussions:) I wanted to make it possible to get all the tasks. What do you think ti make SERIES_ID as required? Just a question, because i think that's appropriate too

src/main/kotlin/com/epam/brn/repo/ExerciseRepository.kt
```diff
@@ -1,8 +1,10 @@
...
 
 import com.epam.brn.model.Exercise
-import org.springframework.data.repository.CrudRepository
+import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.stereotype.Repository
 
 @Repository
-interface ExerciseRepository : CrudRepository<Exercise, Long>
\ No newline at end of file
+interface ExerciseRepository : JpaRepository<Exercise, Long> {
```
goopnigoop: Like it, maybe change it in whole project?

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -24,14 +24,22 @@ class ExerciseService(
...
+    fun findExerciseEntityById(id: Long): Exercise {
         return exerciseRepository.findById(id)
             .orElseThrow { EntityNotFoundException("Exercise entity was not found by id $id") }
     }
+
+    fun findExercisesByUserId(userId: Long): List<ExerciseDto> {
+        log.debug("Searching available exercises for user=$userId")
+        val exercisesIdList = studyHistoryRepository.getDoneExercisesIdList(userId)
+        val history = exerciseRepository.findAll()
+        return emptyIfNull(history).map { x -> x.toDtoWithoutTasks(exercisesIdList.contains(x.id)) }
```
goopnigoop: I think i missed this part and i don't see the whole picture, but a want to ask:) It Do we have flag available and this flag as i understand shows the availability of the exercise for the current user. Why we are using this flag instead of filter exercises and return only exercises for the current user and get rid of the flag?
As i said maybe i don't see the whole picture

goopnigoop: I've deleted previous comment, i think i get it:)

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
goopnigoop: Thank you, @rpcep

src/main/kotlin/com/epam/brn/repo/ExerciseRepository.kt
goopnigoop: me too

src/main/kotlin/com/epam/brn/config/DevConfig.kt
goopnigoop: I don't like LoadHandBook for init loading the same data on all kinds of environments without a configuration.
Furthermore, LoadHandBook was started during tests and put data in the embedded database, and I got something inside the database at the beginning of the test
I'm not sure about configuration honestly, but i definitely want to define a profile.
Anyway, later we will use a flyway with versioning scripts to init database. And I'm waiting to implement it.

goopnigoop: And it's good to approach to fill database in ApplicationRunner implementation

goopnigoop: As for me, I preferred to listen ContextRefreshedEvent and init what i want, but this approach I used in spring and didn't use services. That's why i suppose ApplicationRunner is good approach from my point of view

ElenaSpb:
lets leave LoadHandBook for loading initial data as temporary step as it was done.
(prod is not happen tommorrow)
second step of plan was to load initial data from file. 
but not in configuration. 
lets use configuration for configuration. (not for loading data, please).

in context listener we can check (in another task we will do it) with what profile application is started - in test / prod envirnment lets not load initial data - not call the function loadInitialDataToDb()

src/main/java/com/epam/brn/service/LevelCalculationService.java
```diff
@@ -1,7 +0,0 @@
-package com.epam.brn.service;
-
-import org.springframework.stereotype.Service;
-
-@Service
```
ElenaSpb: lets stay this service as an example/template as it was done
that java code also can be in project

src/main/kotlin/com/epam/brn/config/DevConfig.kt
```diff
@@ -33,11 +35,12 @@ class DevConfig(
...
     fun databaseInitializer() = ApplicationRunner {
         log.debug("------- Started DEV initialization")
-        createDBMockContentValues()
+        loadInitialDBDataForAPIDev()
+        loadInitialDBDataForFileUploadDev()
         log.debug("------- Finished DEV initialization")
     }
 
-    private fun createDBMockContentValues() {
+    private fun loadInitialDBDataForAPIDev() {
```
ElenaSpb: lets move all initial loading logic to separate class - LoadHandBook
from config

src/main/kotlin/com/epam/brn/config/DevConfig.kt
rpcep: Apparently we will not be listened to

src/main/kotlin/com/epam/brn/config/DevConfig.kt
rpcep: Done.

src/main/java/com/epam/brn/service/LevelCalculationService.java
rpcep: Done.

src/main/kotlin/com/epam/brn/model/Exercise.kt
```diff
@@ -31,18 +31,21 @@ data class Exercise(
...
     val tasks: MutableSet<Task> = HashSet()
 ) {
-    fun toDtoWithTasks() = ExerciseDto(
+    fun toDtoWithTasks(available: Boolean? = null) = ExerciseDto(
         seriesId = series.id,
         id = id,
         name = name,
         description = description,
-        tasks = tasks.map { task -> task.toDto() }.toMutableSet()
+        tasks = tasks.map { task -> task.toDto() }.toMutableSet(),
```
ElenaSpb: as we agreed to return entity with child ids by default - so we need there task.id

rpcep: That is another task  "EPMLABSBRN-106".
This is "EPMLABSBRN-101".

ElenaSpb: https://jira.epam.com/jira/browse/EPMLABSBRN-106 is more a bug that now it does not return anything.
but there we see, that accoding raml we need not now both of this functions

rpcep: That jira issue is classified as a task, and says "GET end-point by default contains child ids as array. https://kb.epam.com/display/EPMLABSBRN/API+Contract " which is what you are requesting.

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
ElenaSpb: logically exercises belong to series so **seriesId is required parameter** for end-point like `/exercises'
also for calculation field `available` we need userId (as we have not reg/login module yet, we can use userId=1 by default)

rpcep: I do not agree.

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -22,10 +23,15 @@ class ExerciseController(@Autowired val exerciseService: ExerciseService) {
...
     @GetMapping
     @ApiOperation("Get done exercises for user")
-    fun getExercisesByUserID(
-        @RequestParam(value = USER_ID) userId: Long
+    fun getExercises(
+        @RequestParam(value = USER_ID, required = true) userId: Long,
+        @RequestParam(value = SERIES_ID, required = true) seriesId: Long?
     ): ResponseEntity<BaseResponseDto> {
-        return ResponseEntity.ok().body(BaseResponseDto(data = exerciseService.findDoneExercisesByUserId(userId)))
+        seriesId?.let {
```
ElenaSpb: so  we need not additional check with let here as
seriesId is required

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -22,10 +23,15 @@ class ExerciseController(@Autowired val exerciseService: ExerciseService) {
...
+            return ResponseEntity.ok()
+                .body(BaseResponseDto(data = exerciseService.findExercisesByUserIdAndSeries(userId, it)))
+        }
+        return ResponseEntity.ok().body(BaseResponseDto(data = exerciseService.findExercisesByUserId(userId)))
```
ElenaSpb: and please remove this one line

PR #39, opened 2019-11-07
Epmlabsbrn 113 "errors" filed in API responce

ElenaSpb:
EPMLABSBRN-113 "errors" filed in API responce

fix format of Api response,
fix tests


ElenaSpb:
> Could you please merge the last task 101 before creating changes that modify it?

yes

PR #41, opened 2019-11-07
EPMLABSBRN-114 Back-end: load initial data only in dev/prod profile

ElenaSpb:
EPMLABSBRN-114 Back-end: load initial data only in dev/prod profile

add support profiles active for load initial data


ElenaSpb:
> The profile is fine.

Ruben, thanks for ideas,
but as I checked it - it was really called once in debug.

PR #58, opened 2019-11-13
[EPMLABSBRN-129] - fix audio_file_url setting in load job

RiveriRiv:
[EPMLABSBRN-129](https://jira.epam.com/jira/browse/EPMLABSBRN-129)

- Fixes audio_file_url setting in load job

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
+++ b/src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
@@ -1,6 +1,5 @@
 package com.epam.brn.service.parsers.csv.converter.impl
 
-import com.epam.brn.exception.EntityNotFoundException
 import com.epam.brn.model.Resource
 import com.epam.brn.model.Task
 import com.epam.brn.service.ExerciseService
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
```diff
@@ -55,24 +54,25 @@ class TaskCsvToTaskModelConverter : Converter<TaskCsv, Task> {
...
 
         return resources[0]
     }
 
-    private fun createResource(word: String, fileName: String): Resource {
+    private fun createResource(word: String): Resource {
         val resource = Resource()
         resource.word = word
-        resource.audioFileUrl = fileName
+        resource.audioFileUrl = "$word.mp3"
```
goopnigoop: please put it to constants or maybe in properties but not here

ElenaSpb: as $word is dynamic part (where word is the name of variable)
so logically move only const only ".mp3" part

ElenaSpb:
exception is linked to the task

src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
ElenaSpb: this need to be set after check that there are no already such resource in db

ElenaSpb: so we need to create resource for word in answer array if there are no already resource with such word in db.
we need to create resource for correct word with its filename which is set in file.

PR #59, opened 2019-11-14
EPMLABSBRN-110 - Added raml description for tasks endpoint

goopnigoop:
Raml documentation is updated with tasks endpoint


api-contract/api.raml
```diff
@@ -70,16 +70,27 @@ mediaType:
 
 /tasks:
   get:
+    description: |
+    The tasks endpoint returns tasks assotiated with specified exercise id
     queryParameters:
       exerciseID:
```
ElenaSpb: may be it should be `exerciseId`

goopnigoop: ok

goopnigoop:
Changes after review are done

PR #72, opened 2019-11-18
EPMLABSBRN-78

ElenaSpb:
EPMLABSBRN-78
add spring cache - caffeine

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/TaskService.kt
+++ b/src/main/kotlin/com/epam/brn/service/TaskService.kt
@@ -5,13 +5,18 @@ import com.epam.brn.exception.NoDataFoundException
 import com.epam.brn.model.Task
 import com.epam.brn.repo.TaskRepository
 import org.apache.logging.log4j.kotlin.logger
+import org.springframework.cache.annotation.CacheConfig
+import org.springframework.cache.annotation.CacheEvict
+import org.springframework.cache.annotation.Cacheable
 import org.springframework.stereotype.Service
 
 @Service
+@CacheConfig(cacheNames = arrayOf("tasks"))
 class TaskService(private val taskRepository: TaskRepository) {
 
     private val log = logger()
 
+    @Cacheable
     fun getAllTasksByExerciseId(exerciseId: Long): List<TaskDto> {
         val tasks = taskRepository.findAllTasksByExerciseIdWithJoinedAnswers(exerciseId)
         return tasks.map { task -> task.toDto() }
@@ -26,6 +31,7 @@ class TaskService(private val taskRepository: TaskRepository) {
         return taskRepository.save(task)
     }
 
+    @CacheEvict("tasks")
     fun getTaskById(taskId: Long): TaskDto {
         log.debug("Searching task with id=$taskId")
         return taskRepository.findById(taskId)
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -21,3 +21,6 @@ brn.task.files.path=/inputData/tasks
 brn.processed.files.path=/inputData/processed
 
 server.servlet.context-path=/api
+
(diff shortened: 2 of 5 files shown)


src/main/resources/application.properties
```diff
@@ -20,4 +20,7 @@ spring.flyway.enabled=false
 brn.task.files.path=/inputData/tasks
 brn.processed.files.path=/inputData/processed
 
-server.servlet.context-path=/api
\ No newline at end of file
+server.servlet.context-path=/api
+
+spring.cache.cache-names=seriesByGroupId, seriesById, tasks
+spring.cache..caffeine.spec=maximumSize=1000, expireAfterAccess=3600s
```
goopnigoop: 2 points in prop. name

src/main/kotlin/com/epam/brn/service/TaskService.kt
```diff
@@ -26,6 +31,7 @@ class TaskService(private val taskRepository: TaskRepository) {
         return taskRepository.save(task)
     }
 
+    @CacheEvict("tasks")
```
goopnigoop: Don't understand why do use CacheEvict here?

src/main/resources/application.properties
ElenaSpb: yes
it`s caffeine

src/main/kotlin/com/epam/brn/service/TaskService.kt
ElenaSpb: if task is updated while loading

goopnigoop: why there is no cachable?

src/main/resources/application.properties
goopnigoop: tell me please, why 2 points?

goopnigoop: I think it's a typo, haven't seen 2 points

rpcep: There is an extra dot in the property.
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-caching-provider-caffeine

src/main/kotlin/com/epam/brn/service/TaskService.kt
```diff
@@ -5,13 +5,18 @@ import com.epam.brn.exception.NoDataFoundException
...
+import org.springframework.cache.annotation.Cacheable
 import org.springframework.stereotype.Service
 
 @Service
+@CacheConfig(cacheNames = arrayOf("tasks"))
 class TaskService(private val taskRepository: TaskRepository) {
 
     private val log = logger()
 
+    @Cacheable
```
rpcep: Joined answers may change here, should we use CacheEvict?

src/main/kotlin/com/epam/brn/service/TaskService.kt
rpcep: We do not join answers in this method, in which case it may be updated while loading? Would it be more likely than updating a series while loading? We do cache those.

src/main/resources/application.properties
ElenaSpb: yes, it is misprint

src/main/kotlin/com/epam/brn/service/TaskService.kt
ElenaSpb: we load tasks, not series.
so tasks can be changed

src/main/kotlin/com/epam/brn/service/TaskService.kt
ElenaSpb: no changes should be in getFunction

ElenaSpb:
now I add some configuration for caffein..

Mikhail-Sosnin:
@elainecode please, take care of the merge conflicts.

PR #75, opened 2019-11-19
EPMLABSBRN-128 add initial data and files

ElenaSpb:
EPMLABSBRN-128 
add initial data and files


ElenaSpb:
it is just preporation to https://github.com/Brain-up/brn/issues/73
where hard code would be removed

src/main/kotlin/com/epam/brn/service/LoadHandBook.kt
```diff
@@ -47,60 +47,92 @@ class LoadHandBook(private val exerciseGroupRepository: ExerciseGroupRepository)
...
+        )
+        val exercise3 = Exercise(
+            name = "Однослоговые слова без шума",
+            description = "Однослоговые слова без шума",
+            level = 2,
+            series = series1
+        )
+
+        val exercise4 = Exercise(
+            name = "Однослоговые слова c малым шумом",
```
zinch84: Better to extract duplicating literals to a variable

ElenaSpb: it is only a temporary inserting, it will be removed soon, you know )

PR #80, opened 2019-11-21
EPMLABSBRN-142 exercises completion state

iki6:
[EPMLABSBRN-142](https://jira.epam.com/jira/browse/EPMLABSBRN-142)

**Description**: exercises completion state


frontend/app/routes/series/exercise.js
```diff
@@ -10,10 +10,13 @@ export default checkInteractionRoute.extend({
...
     if (
       to.name.endsWith('exercise.index') &&
-      exercise.tasks.firstObject &&
+      exercise.sortedChildren.firstObject &&
       !to.paramNames.includes('task_id')
     ) {
-      this.transitionTo('series.exercise.task', exercise.tasks.firstObject);
+      this.transitionTo(
+        'series.exercise.task',
+        exercise.sortedChildren.firstObject,
```
H1D: 1) It is better to have readable names: `exercise.sortedChildren.firstObject` => `exercise.sortedTasks.firstObject`
2) Let's always pass id's to `transitionTo()` and `link-to`. Otherwise router will skip `model()` invocation which is bad for testability and consistency because sometimes it is called and sometimes not.

PR #92, opened 2019-12-03
Epmlabsbrn-137 devops: create build process for 3 docker images

DmitriyStoyanov:
[EPMLABSBRN-137](https://jira.epam.com/jira/browse/EPMLABSBRN-137)

**Description**:
devops: create build process for 3 docker images

- Decreased size of docker images for frontend and backend
- Created sample of build process on github actions
- Configured build process on dockerhub with next images:
* https://hub.docker.com/r/brainup/brn-frontend
* https://hub.docker.com/r/brainup/brn-backend
- Created docker-compose-run to run application without build process


Mikhail-Sosnin:
Nicely done!
@DmitriyStoyanov  Could you please take care of the commits count via squash?

Mikhail-Sosnin:
@DmitriyStoyanov is it possible to perform a kind of hardening (https://medium.com/intive-developers/hardening-docker-quick-tips-54ca9c283964) it this request or it would be better to create separate task? From the security perspective it would be perfect if the containers might be launched with the limited permissions.

DmitriyStoyanov:
@Mikhail-Sosnin described article more regarding host hardening, only change to image we can add is separate specific user. But i guess it is a story about running containers which can be done in separate task after deployment process will be created

ElenaSpb:
Dmitriy, do we need to add or change something in our Readme file?

ElenaSpb:
> @Mikhail-Sosnin described article more regarding host hardening, only change to image we can add is separate specific user. But i guess it is a story about running containers which can be done in separate task after deployment process will be created

so could you create a task for it to not forget about this point?

PR #111, opened 2019-12-19
EPMLABSBRN-158  studying timer

iki6:
[EPMLABSBRN-158](https://jira.epam.com/jira/browse/EPMLABSBRN-158)

**Description**: studying timer


frontend/app/components/timer/component.js
```diff
@@ -0,0 +1,47 @@
...
+    const seconds = this.countedSeconds % 60;
+    return `${leadingZero(mins)}:${leadingZero(seconds)}`;
+  }),
+  runTimer() {
+    this.set(
+      'timer',
+      setInterval(() => {
+        if (!this.isPaused)
+          this.studyingTimer.incrementProperty('countedSeconds');
+      }, 1000),
```
H1D: 1) never ever use setInterval/setTimeout/timeout (any other form of delay/sleep) to count time. Duration you set there are just suggestion to browser. There is no guarantee you function will be called exactly after 1000ms.
Correct approach would be remembering timer start moment and calculation difference each time.

2) you shouldn't use setInterval/setTimeout with ember. Since ember do not know about those functions being called it may cause some weird bugs. For example during testing ember can stop test execution before setTimeout callback. Use `later`.


```javascript
 const { later, cancel, }

updateSecondsCount() {
   const secondsPassed = (new Date) - this.timerStart;
   this.set('countedSeconds', secondsPassed); 
   this.set('timerId',later(this,this.updateSecondsCount,1000));
}

startTimer() {
  this.set('timeStart',(new Date).getTime());
  this.set('timerId',later(this,this.updateSecondsCount,1000));
}


stopTimer() {
   clearInterval(this.timerId);
   this.set('timeStart', null);
}
```

PR #116, opened 2019-12-30
EPMLABSBRN-82

iki6:
[EPMLABSBRN-82](https://jira.epam.com/jira/browse/EPMLABSBRN-82)

**Description**: show words sequences tasks


frontend/app/models/task/single-words.js
```diff
@@ -0,0 +1,10 @@
+import DS from 'ember-data';
+const { attr } = DS;
+import BaseTask from '../task';
+
+export default class SingleWords extends BaseTask.extend({
```
H1D: one task represents single word, so better naming would be `SingleWordTask`

PR #183, opened 2020-01-20
EPMLABSBRN-207 timer autopause

iki6:
[EPMLABSBRN-207](https://jira.epam.com/jira/browse/EPMLABSBRN-207)

**Description**: timer autopause

### DIFF
--- a/frontend/tests/integration/components/timer/component-test.js
+++ b/frontend/tests/integration/components/timer/component-test.js
@@ -3,7 +3,6 @@ import { setupRenderingTest } from 'ember-qunit';
 import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import customTimeout from 'brn/utils/custom-timeout';
-import TimerService from 'brn/services/studying-timer';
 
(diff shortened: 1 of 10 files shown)


frontend/tests/integration/components/timer/component-test.js
```diff
@@ -29,4 +31,25 @@ module('Integration | Component | timer', function(hooks) {
...
+    await render(hbs`<Timer @studyingTimer={{this.studyingTimer}}/>`);
+
+    assert.dom('[data-test-timer-display-value]').hasText('00:00');
+
+    assert
+      .dom('[data-test-timer-wrapper]')
+      .hasNoAttribute('data-test-timer-is-paused');
+
+    await customTimeout();
+    await customTimeout();
```
H1D: earlier `idleTime` is set to `10ms`. But `await customTimeout()` called only twice which will wait only for 2ms while in `'test'` environment accroding to code of ` customTimeout` – https://github.com/Brain-up/brn/blob/dev/frontend/app/utils/custom-timeout.js#L5

PR #194, opened 2020-01-22
EPMLABSBRN-176 add SonarCloud integration

Mikhail-Sosnin:
[EPMLABSBRN-176](https://jira.epam.com/jira/browse/EPMLABSBRN-176)

**Description**: SonarCloud integration.
Here we could find the analysis results: https://sonarcloud.io/dashboard?id=Brain-up_brn

Documentation:
- https://docs.sonarqube.org/latest/analysis/languages/java/
- https://docs.sonarqube.org/latest/


ElenaSpb:
could you please add a section in readme file about sonar work and how every developer can check it report before/after MR

sonar-project.properties
```diff
@@ -0,0 +1,8 @@
+sonar.organization=Brain-up
+sonar.projectKey=Brain-up
+
+sonar.sources=src/main/java
```
goopnigoop: From my point of view, we have to use here source path to kotlin sources, unfortunately, i have no idea whether sonar has rules for kotlin or not

goopnigoop:
Like that now we have sonar check

DmitriyStoyanov:
Looks like sonarcloud trigger failed because of:

Gradle project detected. You should use the SonarQube plugin for Gradle during build rather than using this GitHub Action.
##[error]Docker run failed with exit code 1

sonar-project.properties
Mikhail-Sosnin: updated

PR #195, opened 2020-01-22
EPMLABSBRN-178 npm audit integration

Mikhail-Sosnin:
[EPMLABSBRN-178](https://jira.epam.com/jira/browse/EPMLABSBRN-178)

**Description**: NPM Audit scanner integration

### DIFF
--- a/.github/workflows/fe-composition-analysis.yml
+++ b/.github/workflows/fe-composition-analysis.yml
@@ -1,0 +1,38 @@
+name: Frontend Composition Analysis
+
+on:
+  pull_request:
+    paths:
+      - 'Dockerfile_frontend'
+      - 'frontend/**'
+      - '.github/workflows/fe-composition-analysis.yml'
+    branches:
+      - master
+  push:
+    paths:
+      - 'Dockerfile_frontend'
+      - 'frontend/**'
+      - '.github/workflows/fe-composition-analysis.yml'
+    branches:
+      - master
+
+jobs:
+  npm_audit_scan:
+    name: npm audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: rwjblue/setup-volta@v1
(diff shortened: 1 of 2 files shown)


Mikhail-Sosnin:
> Sorry, I dont like this idea for two reasons:
> 
> 1. It may block PR at a random moment when new vulnerabilities has been [published](https://www.npmjs.com/advisories). And it won't imporve security or anything else. Becasue those new vulnerabilities most probably will be related dependancies already used in `master`. Ideally it should test only newly added depedancies but there is no such tool
> 2. (minor) We use `yarn` and it is better to use it in CI as well.
> 
> My suggestions:
> 
> 1. Agree on some repetitive routine task (have it in JIRA). For example run `yarn audit --level critical` and resolve issues
> 2. (worse) add `yarn audit --level critical` as precommit hook.

@H1D Thanks for the review and valuable comments. So, let me a bit clarify the suggested updates.
Reasons:
1. Yes, it should block PR in case of the new vulnerability identification. This is one of the simplest ways to keep the application secure. As a project team we are responsible for the vulnerabilities handling, especially in the **master** branch. In case we have insecure dependency we automatically have a build-in vulnerabilities which might allow an attacker to simply exploit them and as a result make a negative impact on the product and users. Code analysis tools usage is a best practice and having it on the CI/CD side leads to the pretty high quality and security levels. As for the dependencies findings, in common case this might be fixed via simple **npm audit fix** call.

2. Agree

Suggestions:
1. This might leads to the situation when nobody will do it due to the probable low motivation or team expertise level. Scan needs to be added to the CI/CD part. 
2. Agree with criticals, @ElenaSpb what about vulnerabilities with **high** or **medium** severity?

Mikhail-Sosnin:
@H1D @ElenaSpb updated, please, let me know in case we need an additional configuration.

ElenaSpb:
> 2\. **medium**

medium looks like more reasonable for our project

PR #206, opened 2020-01-23
Epmlabsbrn-204 - exercises opening(becoming available) order

iki6:
[EPMLABSBRN-204](https://jira.epam.com/jira/browse/EPMLABSBRN-204)

**Description**: exercises availability changes


iki6:
as we discussed with @ElenaSpb 
this PR makes all first-level exercises in the group of exercises with the same name available by default
exercises become available in all name-groups separately as the user passes them

PR #216, opened 2020-01-27
EPMLABSBRN-244 githooks were added to the repository

goopnigoop:
[EPMLABSBRN-244](https://jira.epam.com/jira/browse/EPMLABSBRN-244)

githooks were added to the project:

- [x] **pre-commit** (commit message should have prefix: EPMLABSBRN-X)
- [x] **pre-push** (no push to dev/master)
- [x] **commit-msg** (no commit to dev/master)

- a command to set core.hooksPath was added to Makefile
- core.hooksPath automatically set during the build

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,12 @@
+#!/bin/bash
+master_branches=('origin/master' 'master' 'origin/dev' 'dev')
+current_branch=$(git symbolic-ref HEAD | sed -e 's,.*/\(.*\),\1,')
+
+for branch in "${master_branches[@]}"
+do
+    if [ $branch = $current_branch ]; then
+        echo "You're about to commit to $branch and it's not allowed"
+        exit 1
+    fi
+done
+exit 0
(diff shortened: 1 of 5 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+master_branches=('origin/master' 'master' 'origin/develop' 'develop')
```
ElenaSpb: our current dev branch is
**origin/dev**

.githooks/pre-push
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+protected_branch_master='master'
+protected_branch_dev='dev'
```
ElenaSpb: is `dev` ok?
or it should be `origin/dev`?

.githooks/pre-commit
goopnigoop: thanks, I'll fix it

.githooks/pre-push
goopnigoop: it's ok

PR #243, opened 2020-02-03
EPMLABSBRN-213 add api for 3 series

ElenaSpb:
EPMLABSBRN-213 add api for 3 series


src/main/resources/initFiles/tasks_for_sentence_series.csv
```diff
@@ -0,0 +1,3 @@
+orderNumber words answerAudioFile answerParts
+1; (),(),(девочка, дедушка, бабушка), (бросает, читает, рисует),(),(); series3/девочка_читает.mp3; (девочка, читает)
```
goopnigoop: Are all these spaces necessary here? Maybe better:
1;(),(),(девочка, дедушка, бабушка),(бросает, читает, рисует),(),();series3/девочка_читает.mp3;(девочка, читает)
CSV parsing should have rules and you have to include spaces to the rules

src/main/kotlin/com/epam/brn/model/Task.kt
```diff
@@ -40,7 +41,14 @@ data class Task(
         joinColumns = [JoinColumn(name = "task_id", referencedColumnName = "id")],
         inverseJoinColumns = [JoinColumn(name = "resource_id", referencedColumnName = "id")]
     )
-    var answerOptions: MutableSet<Resource> = hashSetOf()
+    var answerOptions: MutableSet<Resource> = hashSetOf(),
+    @ManyToMany(cascade = [(CascadeType.ALL)])
```
goopnigoop: We have to use only CascadeType.MERGE

ElenaSpb: why

src/main/resources/initFiles/tasks_for_sentence_series.csv
ElenaSpb: с пробелами более читабельно

src/main/kotlin/com/epam/brn/model/Task.kt
goopnigoop: I would say just believe, you can read articles, main idea Tasks are independent of Resources when you dele task you shouldn't delete resources

ElenaSpb: yes, thanks, fixed

PR #248, opened 2020-02-03
EPMLABSBRN-225 create auth api

goopnigoop:
[EPMLABSBRN-225](https://jira.epam.com/jira/browse/EPMLABSBRN-225)

**New Api for User Management**

- [x] add new user  | **_ONLY BY ADMIN_**
- [x] get user by Id | **_ONLY BY ADMIN_**
- [x] get user by username | **_ONLY BY ADMIN_**
- [x] delete user by Id | **_ONLY BY ADMIN_**
- [x] get logged user from the session

**Raml changes**
- [x] entry points are described 

**Services**
- [x] new service to get the current logged in user

**Tests**
- [x] unit tests

### DIFF
--- a/src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
+++ b/src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
@@ -11,7 +11,9 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
@@ -23,16 +25,35 @@ import org.springframework.web.bind.annotation.RestController
 @Api(value = BrnPath.USERS, description = "Contains actions over user details and accounts")
 class UserDetailController(@Autowired val userAccountService: UserAccountService) {
 
-    @GetMapping
-    @ApiOperation("Get user by username")
-    fun findUserByName(@RequestParam(BrnParams.USER_NAME) userName: String): ResponseEntity<BaseResponseDto> {
-        return ResponseEntity.ok()
-            .body(BaseResponseDto(data = listOf(userAccountService.findUserByName(userName))))
+    @PostMapping
+    fun addUser(@Validated @RequestBody userAccountDto: UserAccountDto): ResponseEntity<BaseResponseDto> {
+        val newUser = userAccountService.addUser(userAccountDto)
+        return ResponseEntity.status(HttpStatus.CREATED).body(BaseResponseDto(data = listOf(newUser)))
     }
 
-    @PostMapping
-    fun addUser(@Validated @RequestBody userAccountDto: UserAccountDto): ResponseEntity<UserAccountDto> {
-        // TODO implement adding new user and update com.epam.brn.dto.UserAccountDto
-        return ResponseEntity.status(HttpStatus.CREATED).body(null)
+    @GetMapping(value = ["/{${BrnParams.USER_ID}}"])
+    @ApiOperation("Get user by Id")
+    fun findUserById(@PathVariable(BrnParams.USER_ID) id: Long): ResponseEntity<BaseResponseDto> {
+        return ResponseEntity.ok()
+            .body(BaseResponseDto(data = listOf(userAccountService.findUserById(id))))
     }
+
+    @GetMapping(value = ["/${BrnParams.CURRENT_USER}"])
+    @ApiOperation("Get current logged in user")
+    fun getCurrentUser() = ResponseEntity.ok()
+        .body(BaseResponseDto(data = listOf(userAccountService.getUserFromTheCurrentSession())))
+
+    @DeleteMapping(value = ["/{${BrnParams.USER_ID}}"])
+    @ApiOperation("Delete a user from the system")
+    fun deleteUser(@PathVariable(BrnParams.USER_ID) id: Long): ResponseEntity<BaseResponseDto> {
+        userAccountService.removeUserWithId(id)
+        return ResponseEntity.ok(BaseResponseDto(data = listOf("Successfully deleted")))
+    }
+
+    @GetMapping
+    @ApiOperation("Get user by username")
+    fun findUserByName(
+        @RequestParam(BrnParams.USER_NAME, required = true) userName: String
+    ) = ResponseEntity.ok()
+        .body(BaseResponseDto(data = listOf(userAccountService.findUserByName(userName))))
 }
--- a/src/main/kotlin/com/epam/brn/dto/UserAccountDto.kt
+++ b/src/main/kotlin/com/epam/brn/dto/UserAccountDto.kt
@@ -1,11 +1,33 @@
 package com.epam.brn.dto
 
+import com.epam.brn.model.UserAccount
+import com.fasterxml.jackson.annotation.JsonInclude
 import java.time.LocalDate
+import javax.validation.constraints.Email
+import javax.validation.constraints.NotBlank
+import javax.validation.constraints.Size
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 data class UserAccountDto(
     val id: Long? = null,
+    @field:NotBlank
     val userName: String,
+    @field:NotBlank
+    @field:Email
     val email: String,
-    val active: Boolean,
+    val active: Boolean = true,
+    @field:NotBlank
+    @field:Size(min = 4)
+    var password: String? = null,
     val birthDate: LocalDate? = null
-)
+) {
+    var authorities: MutableSet<String>? = mutableSetOf()
+    fun toModel() = UserAccount(
+        id = id,
+        userName = userName,
+        email = email,
+        password = password,
+        active = active,
+        birthDate = birthDate
+    )
+}
--- a/src/main/kotlin/com/epam/brn/model/Authority.kt
+++ b/src/main/kotlin/com/epam/brn/model/Authority.kt
@@ -1,14 +1,10 @@
 package com.epam.brn.model
 
-import javax.persistence.CascadeType
 import javax.persistence.Column
 import javax.persistence.Entity
-import javax.persistence.ForeignKey
 import javax.persistence.GeneratedValue
 import javax.persistence.GenerationType
 import javax.persistence.Id
-import javax.persistence.JoinColumn
-import javax.persistence.ManyToOne
 import javax.persistence.SequenceGenerator
 
 @Entity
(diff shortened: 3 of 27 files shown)


src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
```diff
@@ -23,16 +25,35 @@ import org.springframework.web.bind.annotation.RestController
 @Api(value = BrnPath.USERS, description = "Contains actions over user details and accounts")
 class UserDetailController(@Autowired val userAccountService: UserAccountService) {
 
-    @GetMapping
-    @ApiOperation("Get user by username")
-    fun findUserByName(@RequestParam(BrnParams.USER_NAME) userName: String): ResponseEntity<BaseResponseDto> {
+    @PostMapping
+    fun addUser(@Validated @RequestBody userAccountDto: UserAccountDto): ResponseEntity<BaseResponseDto> {
+        val addedUser = userAccountService.addUser(userAccountDto)
```
ElenaSpb: maybe `newUser` is shoter

src/main/kotlin/com/epam/brn/repo/AuthorityRepository.kt
```diff
@@ -0,0 +1,11 @@
+package com.epam.brn.repo
+
+import com.epam.brn.model.Authority
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+
+@Repository
+interface AuthorityRepository : JpaRepository<Authority, Long> {
+    fun findAuthorityByAuthorityName(authorityName: String): Authority?
+    fun findAuthoritiesById(id: Long): List<Authority>
```
ElenaSpb: if `id` is unique - why there is a list in result?

src/main/kotlin/com/epam/brn/model/Authority.kt
```diff
@@ -21,13 +17,11 @@ data class Authority(
         allocationSize = 50
     )
     var id: Long? = null,
```
ElenaSpb: as id we think is unique - maybe better to add (unique = true) for it?

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -1,25 +1,102 @@
...
+        val setOfAuthorities = getTheAuthoritySet(userAccountDto)
+        val hashedPassword = getHashedPassword(userAccountDto)
+
+        userAccountDto.password = hashedPassword
+        val userAccount = userAccountDto.toModel()
+        userAccount.authoritySet = setOfAuthorities
+        return userAccountRepository.save(userAccount).toDto()
+    }
+
+    private fun getHashedPassword(userAccountDto: UserAccountDto) = passwordEncoder.encode(userAccountDto.password)
```
ElenaSpb: as I think that in userAccountDto password is already hashed, why we need to encode it additionaly

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -1,25 +1,102 @@
...
+
+    private fun getTheAuthoritySet(userAccountDto: UserAccountDto): MutableSet<Authority> {
+        var authorityNames = userAccountDto.authorities ?: mutableSetOf()
+        if (authorityNames.isEmpty()) {
+            authorityNames = mutableSetOf(AUTH_ROLE_USER)
+        }
```
ElenaSpb: lets not use brackets {} if it is a single line after condition
as it is note in https://github.com/Brain-up/brn/wiki/Code-Style

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -1,25 +1,102 @@
...
+    override fun removeUserWithId(id: Long) {
+        findUserById(id)
+        userAccountRepository.deleteById(id)
+    }
+
+    private fun getNameFromPrincipals(authentication: Authentication): String {
+        val principal = authentication.principal
+        if (principal is UserDetails) {
+            return principal.username
+        }
```
ElenaSpb: lets not use brackets {} if it is a single line after condition

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -1,25 +1,102 @@
...
+        if (principal is Principal) {
+            return principal.name
+        }
```
ElenaSpb: lets not use brackets {} if it is a single line after condition

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
ElenaSpb: could you please write in few words - what goal it resolves?

src/main/kotlin/com/epam/brn/dto/UserAccountDto.kt
```diff
@@ -1,11 +1,59 @@
...
     val userName: String,
+    @field:NotBlank
+    @field:Email
     val email: String,
-    val active: Boolean,
+    val active: Boolean = true,
+    @field:NotBlank
+    @field:Size(min = 4)
+    var password: String? = null,
+    val authorities: MutableSet<String>? = mutableSetOf(),
```
odanc: I see you don't want to use `authorities` field for object's identity, so you override `equals()` and `hashCode()`.
I suggest you move `authorities` field from primary constructor to class body, that way it won't be used in generated methods.

src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
```diff
@@ -23,16 +25,35 @@ import org.springframework.web.bind.annotation.RestController
...
+        return ResponseEntity.status(HttpStatus.CREATED).body(BaseResponseDto(data = listOf(addedUser)))
+    }
+
+    @GetMapping(value = ["/{${BrnParams.USER_ID}}"])
```
odanc: Why use arrays here?

src/main/kotlin/com/epam/brn/model/CustomUserDetails.kt
```diff
@@ -9,41 +9,27 @@ import org.springframework.security.core.userdetails.UserDetails
 class CustomUserDetails(userAccount: UserAccount) : UserDetails {
 
     private val userName: String = userAccount.userName
-    private val password: String = userAccount.password
+    private val password: String? = userAccount.password
     private val active = userAccount.active
     private val authorities: List<GrantedAuthority>
 
     init {
         authorities = CollectionUtils.emptyIfNull(userAccount.authoritySet)
```
odanc: No need to wrap `authoritySet`, it is non-nullable by design.
Even if it is nullable, I suggest using `?.` and `?:` instead of 3rd-party library.

src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
goopnigoop: ok

src/main/kotlin/com/epam/brn/repo/AuthorityRepository.kt
goopnigoop: I'll fix

src/main/kotlin/com/epam/brn/model/Authority.kt
goopnigoop: no need because  @Id is set

 * The mapped column for the primary key of the entity is assumed 
 * to be the primary key of the primary table.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
goopnigoop: in few words: to mock final classes

there is much info in google

src/main/kotlin/com/epam/brn/dto/UserAccountDto.kt
goopnigoop: ok, we can do it

src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
goopnigoop: From GetMapping annotation: There is array
From GetMapping annotation

src/main/kotlin/com/epam/brn/model/CustomUserDetails.kt
goopnigoop: ok. thanks

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
ElenaSpb: I meen write a comment before this flag in file.
it would be useful, readable and good pratice before each such flag have a short comment what goal it resolves

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -1,25 +1,102 @@
...
+
+    private fun getNameFromPrincipals(authentication: Authentication): String {
+        val principal = authentication.principal
+        if (principal is UserDetails)
+            return principal.username
+
+        if (principal is Principal)
+            return principal.name
+
+        throw NoDataFoundException("There is no user in the session")
```
ElenaSpb: as we assume that fun need to be shoter lets not make it longer with additional empty lines, please

src/test/kotlin/com/epam/brn/controller/UserDetailControllerTest.kt
```diff
@@ -17,21 +25,79 @@ internal class UserDetailControllerTest {
...
+        fun `should get logged in user from the current session`() {
+            // GIVEN
+            Mockito.`when`(userAccountService.getUserFromTheCurrentSession()).thenReturn(userAccountDto)
+            // WHEN
+            @Suppress("UNCHECKED_CAST")
+            val savedUserAccountDto = userDetailController.getCurrentUser().body?.data as List<UserAccountDto>
+            // THEN
+            Assertions.assertThat(savedUserAccountDto).hasSize(INTEGER_ONE)
+            Assertions.assertThat(savedUserAccountDto[0]).isEqualTo(userAccountDto)
+            verify(userAccountService).getUserFromTheCurrentSession()
```
ElenaSpb: lets import Assertions. and Mockito. as static and do not write it in code.
just 
`when`(..)
assertThat(..)

src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
```diff
@@ -1,39 +1,131 @@
...
+    @Nested
+    @DisplayName("Tests for getting users")
+    inner class GetUserAccounts {
+        @Test
+        fun `should find a user by id`() {
+            // GIVEN
+            val userName = "Tested"
+            `when`(userAccount.toDto()).thenReturn(userAccountDto)
+            `when`(userAccountDto.userName).thenReturn(userName)
+
```
ElenaSpb: please, avoid empty lines

goopnigoop: ok

src/test/kotlin/com/epam/brn/controller/UserDetailControllerTest.kt
goopnigoop: ok

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
goopnigoop: ok

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
goopnigoop: It's not hashed, when a user creates user it writes the password and then it will be hashed

ElenaSpb: we get password from front in hashed state already. not?

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
ElenaSpb: yes, first block it quit short and readable

PR #255, opened 2020-02-06
EPMLABSBRN-279 fix wordType in resources

ElenaSpb:
EPMLABSBRN-279 fix wordType in resources

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
+++ b/src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
@@ -1,5 +1,6 @@
 package com.epam.brn.service.parsers.csv.converter.impl
 
+import com.epam.brn.constant.WordTypeEnum
 import com.epam.brn.exception.EntityNotFoundException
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/epam/brn/service/parsers/csv/converter/impl/TaskCsvToTaskModelConverter.kt
```diff
@@ -78,19 +79,19 @@ class TaskCsvToTaskModelConverter : Converter<TaskCsv, Task> {
     private fun getResourceByWord(word: String): Resource {
         val resources = resourceService.findByWordLike(word)
         return if (CollectionUtils.isEmpty(resources))
-            createAndGetResource(word, StringUtils.EMPTY, StringUtils.EMPTY)
+            createAndGetResource(word, StringUtils.EMPTY, StringUtils.EMPTY, WordTypeEnum.OBJECT.toString())
```
RiveriRiv: why does wordType always OBJECT, could we get this value from source.wordType?

ElenaSpb: it is in case that there are no resource in db and no any additional information about it

PR #313, opened 2020-02-19
EPMLABSBRN-287 rest end-points for Basic auth

ElenaSpb:
EPMLABSBRN-287 rest end-point for auth
add

[EPMLABSBRN-287](https://jira.epam.com/jira/browse/EPMLABSBRN-287)


src/main/kotlin/com/epam/brn/constant/BrnPath.kt
```diff
@@ -2,6 +2,9 @@ package com.epam.brn.constant
 
 object BrnPath {
     const val HOME = "/home"
+    const val API = "/api"
+    const val AUTH = "/oauth"
```
lifeart: should it be `/ouath/token`?

ElenaSpb: yes, finall yit would be `/ouath/token`

api-contract/api.raml
```diff
@@ -161,6 +161,37 @@ mediaType:
...
+/brnlogin:
+  post:
+  description: Login for exist user
+    body:
+     application/json:
+        example: !include samples/login.json
+    responses:
+      200:
+        example: !include samples/authSuccessResponse.json
+      403:
```
goopnigoop: In case you return bad credentials, we have to use **401** instead of 403, because you want to say user that it is unauthorized

api-contract/api.raml
```diff
@@ -161,6 +161,37 @@ mediaType:
...
+        example: !include samples/authBadCredResponse.json
+
+/registration:
+  post:
+  description: Registration for new user with login him
+    body:
+      application/json:
+        example: !include samples/registration.json
+    responses:
+      201:
```
goopnigoop: I don't think we have to return token here, because registration that's enough when the user is created and it's enough.
When a user wants to login it should use login entry point

api-contract/api.raml
```diff
@@ -161,6 +161,37 @@ mediaType:
...
+        example: !include samples/authSuccessResponse.json
+      403:
```
goopnigoop: 403 is not appropriate here, i think we have to return 400 or 409 in case we can not register user because of conflict

goopnigoop:
@lifeart please check my comments in raml, what do you think?

src/main/kotlin/com/epam/brn/controller/AuthenticationController.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping
+@Api(description = "Contains login in actions")
+class AuthenticationController(val authenticationService: AuthenticationService) {
+
+    @PostMapping(BrnPath.REGISTRATION)
+    @ApiOperation("New user registration")
+    fun registration(@Validated @RequestBody userAccountDto: UserAccountDto): ResponseEntity<AuthOutDto> {
```
goopnigoop: If you decided to move registration rat here, please delete it from here:
`com.epam.brn.controller.UserDetailController#addUser`

src/main/kotlin/com/epam/brn/config/WebSecurityBasicConfiguration.kt
```diff
@@ -29,21 +33,23 @@ class WebSecurityConfiguration(
...
     }
 
     @Bean
-    fun passwordEncoder(): PasswordEncoder {
-        return BCryptPasswordEncoder()
-    }
+    fun passwordEncoder(): PasswordEncoder = BCryptPasswordEncoder()
+
+    @Bean
+    fun brnAuthenticationManager(): AuthenticationManager? = authenticationManager()
```
goopnigoop: This is solution from javadoc:
org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter#authenticationManagerBean
/**
	 * Override this method to expose the {@link AuthenticationManager} from
	 * {@link #configure(AuthenticationManagerBuilder)} to be exposed as a Bean. For
	 * example:
	 *
	 * <pre>
	 * &#064;Bean(name name="myAuthenticationManager")
	 * &#064;Override
	 * public AuthenticationManager authenticationManagerBean() throws Exception {
	 *     return super.authenticationManagerBean();
	 * }
	 * </pre>
	 *
	 * @return the {@link AuthenticationManager}
	 * @throws Exception
	 */

goopnigoop: More info:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#authenticationmanager-bean

src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
```diff
@@ -66,4 +73,9 @@ class ExceptionControllerAdvice {
         .status(HttpStatus.INTERNAL_SERVER_ERROR)
         .contentType(MediaType.APPLICATION_JSON)
         .body(BaseResponseDto(errors = listOf(e.message.toString())))
+
+    fun makeForbiddenErrorResponseEntity(e: Throwable) = ResponseEntity
+        .status(HttpStatus.FORBIDDEN)
```
goopnigoop: As i described in Raml FORBIDDEN (403) when you have no rights, 401 - when you unauthorized, please change it

src/main/kotlin/com/epam/brn/dto/AuthOutDto.kt
```diff
@@ -0,0 +1,10 @@
+package com.epam.brn.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude
+data class AuthOutDto(
+    val access_token: String,
+    val token_type: String = "Barear",
```
goopnigoop: wrong word :**Barear**

Right word: **Bearer**

src/main/kotlin/com/epam/brn/service/impl/AuthenticationBasicServiceImpl.kt
```diff
@@ -0,0 +1,38 @@
...
+        val newUser = userAccountService.addUser(userAccountDto)
+        log.info("created new user id=${newUser.id}")
+        return login(LoginDto(username = userAccountDto.email, password = userAccountDto.password))
+    }
+
+    override fun login(loginDto: LoginDto): String {
+        val token = UsernamePasswordAuthenticationToken(loginDto.username, loginDto.password)
+        val auth: Authentication = authenticationManager.authenticate(token)
+        SecurityContextHolder.getContext().authentication = auth
+        val basicHeader = getBasicHeader(loginDto.username, loginDto.password)
```
goopnigoop: you can inline variable and return getBasicHeader(loginDto.username, loginDto.password)

api-contract/api.raml
ElenaSpb: I think 403 Forbidden more resonable there, because user try to login with wrong creds -> 
so Bad Credentials and 403,
Alex, what you think about it?

api-contract/api.raml
ElenaSpb: I think it is comfortable and predictable for user when he registrate - then login him after success registration.

api-contract/api.raml
ElenaSpb: fixed for 400 (Bad Request) here

src/main/kotlin/com/epam/brn/config/WebSecurityBasicConfiguration.kt
ElenaSpb: nice, thank you!
fixed

src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
ElenaSpb: I think for forbidden (login with wrong creds) it is logical to return 403 FORBIDDEN

src/main/kotlin/com/epam/brn/dto/AuthOutDto.kt
ElenaSpb: fixed, thanks

src/main/kotlin/com/epam/brn/controller/AuthenticationController.kt
ElenaSpb: as we discussed with Alex - old end-point may be useful for add users by admin in some cases

src/main/kotlin/com/epam/brn/service/impl/AuthenticationBasicServiceImpl.kt
ElenaSpb: yes, done, thanks

api-contract/api.raml
goopnigoop: There is definitely no way for 403 because you mentioned with 403 that some data is forbidden fro the user and when he came with the wrong password 403 confused him.

api-contract/api.raml
goopnigoop: You can see other sites, just imagine are you really logged in after your registration?:)

ElenaSpb: yes

goopnigoop: please, give me examples

PR #359, opened 2020-03-10
Epmlabsbrn 306

saim-doruklu:
[EPMLABSBRN-306](https://jira.epam.com/jira/browse/EPMLABSBRN-XX)

**Description**: Refactor initial data loader code to make adding new parsers easier


goopnigoop:
very good PR

src/main/kotlin/com/epam/brn/constant/BrnInitFiles.kt
```diff
@@ -0,0 +1,31 @@
+package com.epam.brn.constant
+
+enum class BrnInitFiles {
+    GROUPS("groups.csv"),
+    SERIES("series.csv"),
+    EXERCISES("exercises.csv"),
```
ElenaSpb: we need not "exercises.csv" at all

src/main/kotlin/com/epam/brn/csv/UploadFromCsvService.kt
```diff
@@ -21,5 +21,5 @@ interface UploadFromCsvService {
      * @return failed csv-lines with errors
      */
     @Throws(FileFormatException::class)
-    fun loadTaskFile(file: File): Map<String, String>
+    fun loadTaskFile(file: File): Map<String, String>?
```
ElenaSpb: why it may be null?
it is better to return empty map then null

ElenaSpb:
now tests fail in branch

src/main/kotlin/com/epam/brn/csv/converter/FileNameToUploader.kt
```diff
@@ -0,0 +1,7 @@
+package com.epam.brn.csv.converter
+
+import com.epam.brn.constant.BrnInitFiles
+
+interface FileNameToUploader {
+    fun getUploaderFor(initFile: BrnInitFiles): Uploader<*, *>?
```
ElenaSpb: why * instead of Any, does they better, why?

src/test/resources/application.properties
```diff
@@ -23,7 +23,6 @@ brn.audio.file.second.series.path=series2/%s.mp3
 brn.picture.file.default.path=pictures/withWord/%s.jpg
 brn.dataFormatNumLines=5
 
-cloud.provider=aws
```
ElenaSpb: why you delete it?
in this task?

src/main/kotlin/com/epam/brn/csv/converter/FileNameToUploader.kt
odanc: Agree, it seems `Uploader` interface doesn' set any type bounds (`out` or `in`) and remains invariant.
Looks like no benefits here

src/main/kotlin/com/epam/brn/csv/converter/impl/DefaultRestUploader.kt
```diff
@@ -0,0 +1,36 @@
...
+        val notSavingEntities = mutableMapOf<String, String>()
+
+        entities.forEach {
+            val key = it.key
+            val task = it.value.first
+            try {
+                if (task != null)
+                    uploader.persistEntity(task)
+                else
+                    it.value.second?.let { errorMessage -> notSavingEntities[key] = errorMessage }
```
mmelnik: Why we care about unsaved tasks here? Shouldn't they be treated earlier as we already know which of them contains null taks? We can filter all nulls for example before even trying to upload them. By the way not sure if I found place where these unsaved values are used.

src/main/kotlin/com/epam/brn/csv/converter/impl/DefaultRestUploader.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private fun <Entity> save(entities: Map<String, Pair<Entity?, String?>>, uploader: Uploader<*, Entity>): Map<String, String> {
+        val notSavingEntities = mutableMapOf<String, String>()
+
+        entities.forEach {
+            val key = it.key
+            val task = it.value.first
+            try {
+                if (task != null)
+                    uploader.persistEntity(task)
```
mmelnik: Class is called Uploader. So why method is called persistEntity() and not uploadCsv() or something like that? Persist usually refers to Repository or Dao layer and means database operation. I advice to avoid such misnaming.

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/ExerciseUploader.kt
```diff
@@ -1,18 +1,45 @@
...
+import com.fasterxml.jackson.databind.ObjectReader
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
 import org.springframework.stereotype.Component
 
 @Component
-class ExerciseCsvConverter : Converter<ExerciseCsv, Exercise> {
+class ExerciseUploader(
+    private val exerciseRepository: ExerciseRepository,
+    private val seriesService: SeriesService
+) : Uploader<ExerciseCsv, Exercise> {
```
mmelnik: This class can be split on 2 at least. I see it does 2 different tasks: persisting and conversion. By the way why it's called Uploader if it's not uploading anything?

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/GroupUploader.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.epam.brn.csv.converter.Uploader
+import com.epam.brn.csv.dto.GroupCsv
+import com.epam.brn.model.ExerciseGroup
+import com.epam.brn.repo.ExerciseGroupRepository
+import com.fasterxml.jackson.databind.ObjectReader
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
+import org.springframework.stereotype.Component
+
+@Component
+class GroupUploader(
```
mmelnik: Looks like misnaming again. File doesn't do any upload if I'm not mistaken, so let's call it according to it's role.

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/GroupUploader.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val exerciseGroupRepository: ExerciseGroupRepository
+) : Uploader<GroupCsv, ExerciseGroup> {
+
+    override fun persistEntity(entity: ExerciseGroup) {
+        exerciseGroupRepository.save(entity)
+    }
+
+    override fun entityComparator(): (ExerciseGroup) -> Int {
```
mmelnik: Isn't it better to make entity comparable? That is 1 of the possible options. Can't say which one is better at the moment, but it's strange to have entity comparator in this Uploader class. That's probably last thing I would expect from this class.

src/main/kotlin/com/epam/brn/csv/converter/impl/secondSeries/SeriesTwoUploader.kt
```diff
@@ -3,37 +3,61 @@ package com.epam.brn.csv.converter.impl.secondSeries
...
     val LEVEL = "level"
     val WORDS = "words"
 
-    @Value(value = "\${brn.audio.file.second.series.path}")
-    private lateinit var audioFileUrl: String
+    override fun persistEntity(entity: Exercise) {
+        exerciseService.save(entity)
+    }
+
+    override fun entityComparator(): (Exercise) -> Int {
```
mmelnik: Again if we need to compare Series we need better place for this comparator.

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/SeriesUploader.kt
```diff
@@ -0,0 +1,50 @@
...
+@Component
+class SeriesUploader(
+    private val seriesRepository: SeriesRepository,
+    private val exerciseGroupsService: ExerciseGroupsService
+) : Uploader<SeriesCsv, Series> {
+    override fun persistEntity(entity: Series) {
+        seriesRepository.save(entity)
+    }
+
+    override fun entityComparator(): (Series) -> Int {
```
mmelnik: Consider better place for this comparator.

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/SeriesOneUploader.kt
```diff
@@ -1,43 +1,72 @@
...
+            .withColumnReordering(true)
+            .withArrayElementSeparator(",")
+            .withHeader()
+
+        return csvMapper
+                .readerWithTypedSchemaFor(TaskCsv::class.java)
+                .with(csvSchema)
+    }
+
+    override fun entityComparator(): (Task) -> Int {
```
mmelnik: Consider better place for this comparator, please.

src/main/kotlin/com/epam/brn/csv/converter/impl/firstSeries/ExerciseUploader.kt
```diff
@@ -1,18 +1,45 @@
...
 
-    @Autowired
-    lateinit var seriesService: SeriesService
+    override fun persistEntity(entity: Exercise) {
+        exerciseRepository.save(entity)
+    }
+
+    override fun entityComparator(): (Exercise) -> Int {
```
mmelnik: Yep, and 3d thing - comparing. That's definitely should be extracted from this class. Because at the moment it violates single responsibility principle.

src/main/kotlin/com/epam/brn/csv/converter/impl/DefaultFileNameToUploader.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Service
+
+@Service
+class DefaultFileNameToUploader(
+    groupUploader: GroupUploader,
+    exerciseUploader: ExerciseUploader,
+    seriesUploader: SeriesUploader,
+    seriesOneUploader: SeriesOneUploader,
+    seriesTwoUploader: SeriesTwoUploader
+) : FileNameToUploader {
```
mmelnik: Isn't there better way to hold all repositories? I'm sure Spring can inject map of objects right in place we need it.

PR #361, opened 2020-03-10
Update repository configuration

mmelnik:
Line endings changes were moved to different branch. These PR contains:

1 EPMLABSBRN-0 update message in commit-msg hook, add message format <eom>

2 EPMLABSBRN-0 remove pre-commit and pre-push git hooks
    
    Current pre-commit and pre-push hooks check user
    active branch and fire alert if it's 'master' or 'dev'.
    This is not conceptually correct as user works with
    his local 'master' and 'dev' copies.
    
    Therefore commit on local 'master' doesn't necesserily means
    that it will be pushed to 'origin/master'. Same for 'dev' branch.
    
    Example:
      git push origing master:some-nice-feature

    Moreover these hooks doesn't prevent me from pushing to 'dev' or 'master'
    if I'm not working on one of these branches.

    Example:
      git push origing some-nice-feature:master     
    
    What we really wanted to achieve here is to prevent pushes
    to 'origin/master' and 'origin/dev'.
   
    And that should be done on server side only. In our case we need
    to make 'dev' and 'master' branches protected in repository settings
    on Github.


.githooks/commit-msg
```diff
@@ -14,8 +13,7 @@ var commitMsg = fs
   .trim();
 
 if (!commit_regex.test(commitMsg)) {
-  process.stderr.write(error_msg);
+  process.stderr.write("Aborting commit. Your commit message is missing valid header or description.\n");
+  process.stderr.write("Use format 'EPMLABSBRN-# issue description' or 'Merge description'. Issue number must be in range [0-1999].\n");
```
ElenaSpb: what the point to make this changes

goopnigoop:
Hi @mmelnik , could you please provide more info in the description?

.githooks/commit-msg
ElenaSpb: and why was deleted pre-commit  and pre-push?

src/main/resources/META-INF/resources/svg/apple.svg
```diff
@@ -1,9 +1,9 @@
...
-</svg>
+<svg
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 100 100"
+    >
+  <title>Apple</title>
+  <g>
+    <ellipse cx="51" cy="51" rx="48" ry="48" fill="#eeee00" stroke="#f0f000" stroke-width="2"/>
+  </g>
+</svg>
```
ElenaSpb: what is the point of this changes?

src/main/resources/META-INF/resources/svg/cylinder.svg
```diff
@@ -1,15 +1,15 @@
...
+  <g>
+    <rect
+        fill="#cc0000"
+        stroke="#cc0000" stroke-width="2"
+        x="1" y="32"
+        width="98" height="38.1"/><!-- 38.1 instead of 38 to ensure overlapping with the bottom arc -->
+    <path d="M1,71 a49,28 0 0,0 98,0" fill="#cc0000" stroke="red" stroke-width="2"/><!-- bottom arc -->
+    <ellipse cx="50" cy="31" rx="49" ry="30" fill="#ee0000" stroke="red" stroke-width="2"/>
+  </g>
+</svg>
```
ElenaSpb: what is the main point of this changes?

.githooks/commit-msg
mmelnik: Commit on local 'master' doesn't necesserily means that it will be pushed to 'origin/master'. Same for 'dev' branch. In this way you force developer to explicitly create additional branches locally that is up to developer and shouldn't be dictader by anyone.

What we really wanted to achieve here is to prevent pushes to 'origin/master' and 'origin/dev' but it cannot be done on client side.

src/main/resources/META-INF/resources/svg/apple.svg
mmelnik: This is re-formatted strings after applying .gitattributes config file. All CRLF line endings were changed to LF.

src/main/resources/META-INF/resources/svg/cylinder.svg
mmelnik: This is re-formatted strings after applying .gitattributes config file. All CRLF line endings were changed to LF.

src/main/resources/META-INF/resources/svg/apple.svg
mmelnik: This is re-formatted strings after applying .gitattributes config file. All CRLF line endings were changed to LF. Because previously some one stored them in CRLF (Win) format.

lifeart:
I don't like idea removing gith hooks because we don't facing described problems

goopnigoop:
> I don't like idea removing gith hooks because we don't facing described problems

I think i agree here

.githooks/commit-msg
mmelnik: Any other thoughts, colleagues?

mmelnik:
> I don't like idea removing gith hooks because we don't facing described problems

I faced this as new developer. And motivation is clearly described. Don't see any value from this hook. And there is no value why we have it in repository?

PR #370, opened 2020-03-11
EPMLABSBRN-0 add .gitattributes file

mmelnik:
.gitattributes file in general controls
line ending on repository level and prevents
related error on different OS without need
of additional tools.

Without it people on different OS very likely will face issues
when IDE/editor automatically change line endings that lead
to full file modification.

It listing binaries files and can be used to set LF/CRLF
endings explicitly for each format.

See https://git-scm.com/docs/gitattributes for details

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,59 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+
+# Common formats
+*.html text
+*.xml text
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,56 @@
...
+*.js text
+*.properties text
+*.bat text
+*.rtf text
+
+# SQL scripts
+*.sql text
+
+# Java sources
+*.java text
```
ElenaSpb: what about Kotlin source ?

mmelnik: Yes, let's add them :)

PR #372, opened 2020-03-12
Epmlabsbrn 311 refactoring aws config

mmelnik:
Described classes are higly coupled that makes
testing of AwsCloudService harder and requires
mocks on data structures.

In general mocking data structures is anti-pattern
and we should avoid such coding, as it signalizes
us that code have smells and have high coupling.

Instead we need to get rid of such hard links
between classes if they are meant to be used
separately.

As simple solution for now is to change Conditions class
from inner to outer and open for world. In order to decouple
it from parent we need to pass all required params
to constructor or builder class.

There are few formatting changes as well.

### DIFF
--- a/src/main/kotlin/com/epam/brn/config/AwsConfig.kt
+++ b/src/main/kotlin/com/epam/brn/config/AwsConfig.kt
@@ -24,8 +24,8 @@ import org.springframework.context.annotation.Profile
 @Profile("!integration-tests")
 @ConditionalOnProperty(name = ["cloud.provider"], havingValue = "aws")
 class AwsConfig(
-    @Value("\${cloud.expireAfterDuration}") expireAfterDuration: String,
-    @Value("\${aws.accessRuleCanned}") accessRuleCanned: String,
+    @Value("\${cloud.expireAfterDuration}") var expireAfterDuration: String,
+    @Value("\${aws.accessRuleCanned}") var accessRuleCanned: String,
     @Value("\${aws.credentialsPath:}") credentialsPath: String,
     @Value("\${aws.accessKeyId:}") accessKeyIdProperty: String,
     @Value("\${aws.secretAccessKey:}") secretAccessKeyProperty: String,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/epam/brn/service/impl/AwsCloudService.kt
```diff
@@ -31,8 +31,7 @@ class AwsCloudService(@Autowired private val awsConfig: AwsConfig) : CloudServic
     override fun bucketUrl(): String = awsConfig.bucketLink
 
     override fun uploadForm(filePath: String): Map<String, Any> {
-        val conditions = awsConfig.getConditions(filePath)
-        return signature(conditions)
+        return signature(awsConfig.buildConditions(filePath))
```
ElenaSpb: it can be one line
_override fun uploadForm(filePath: String): Map<String, Any>= signature(awsConfig.buildConditions(filePath))_

src/main/kotlin/com/epam/brn/config/AwsConfig.kt
```diff
@@ -93,35 +91,56 @@ class AwsConfig(
...
+            amzDateTime, expiration, filePath, successActionRedirect, contentTypeStartsWith, metaTagStartsWith)
     }
 
-    inner class Conditions constructor(now: OffsetDateTime, uuidString: String, filePath: String) {
-        val date: String = dateFormat(now)
-
-        // POLICY CONDITIONS
-        val bucket: Pair<String, String> = "bucket" to bucketName
+    //TODO: we might need extract this class to separate file
+    // and consider usage of Builder pattern instead of big constructor.
```
ElenaSpb: yes, good idea

ElenaSpb:
could you please fix 
Code Smell A 3 Code Smells

ElenaSpb:
@github-actions
Java CI / build (push) Failing after 52s — build

src/main/kotlin/com/epam/brn/service/impl/AwsCloudService.kt
mmelnik: Disagree here is it will be long here and harder to read.
Compare:
    override fun uploadForm(filePath: String): Map<String, Any> {
        return signature(awsConfig.buildConditions(filePath))
    }

        override fun uploadForm(filePath: String): Map<String, Any> = signature(awsConfig.buildConditions(filePath))

ElenaSpb: image

ElenaSpb: this variant is shoter and readable too,
we need not read additional word return and bracets
)

mmelnik: Yep, but you wrote about one line :)

PR #375, opened 2020-03-12
Epmlabsbrn 311 refactoring and issues exception advice

mmelnik:
EPMLABSBRN-311 update log levels
    
    For all cases that are not actual errors from user perspective
    log level was changed from `error` to `warn`
    (i.e. user can upload not valid info, or pass not valid param
    as part of normal program execution).

EPMLABSBRN-311 refactor ExceptionControllerAdvice
    
    - inline methods with 1 usage
    - extract duplicated code to separate method

EPMLABSBRN-311 remove unused code <eom>

### DIFF
--- a/src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
+++ b/src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
@@ -3,8 +3,6 @@ package com.epam.brn.controller.advice
 import com.epam.brn.dto.BaseResponseDto
 import com.epam.brn.exception.EntityNotFoundException
 import com.epam.brn.exception.FileFormatException
-import java.io.IOException
-import java.lang.IllegalArgumentException
 import org.apache.logging.log4j.kotlin.logger
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
@@ -12,6 +10,7 @@ import org.springframework.http.ResponseEntity
 import org.springframework.security.authentication.BadCredentialsException
 import org.springframework.web.bind.annotation.ControllerAdvice
 import org.springframework.web.bind.annotation.ExceptionHandler
+import java.io.IOException
 
 @ControllerAdvice
 class ExceptionControllerAdvice {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
```diff
@@ -20,7 +19,7 @@ class ExceptionControllerAdvice {
 
     @ExceptionHandler(EntityNotFoundException::class)
     fun handleEntityNotFoundException(e: EntityNotFoundException): ResponseEntity<BaseResponseDto> {
-        logger.error("Entity not found exception: ${e.message}", e)
+        logger.info("Entity not found exception: ${e.message}", e)
```
ElenaSpb: please return info -> error in all this places

ElenaSpb: when we will want to change log level - we need to have this message in log. 
error is sutable and resonable I gues

ElenaSpb:
refactor looks fine, thank you

src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
mmelnik: In general error is something that we want do quicly detect in logs. And these specific cases are not errors from app perspective. It can be warn, but if log these as errors they can spam logs and make them less usable.

src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
```diff
@@ -29,7 +28,7 @@ class ExceptionControllerAdvice {
 
     @ExceptionHandler(FileFormatException::class)
     fun handleFileFormatException(e: FileFormatException): ResponseEntity<BaseResponseDto> {
-        logger.error("File format exception: ${e.message}", e)
```
goopnigoop: I'd suggest at least warn here

src/main/kotlin/com/epam/brn/controller/advice/ExceptionControllerAdvice.kt
goopnigoop: I think warn level is a good idea

ElenaSpb: we planned to return to log question later if it would be huge for analize,
now we want to have all this exception in log as error, or maybe warn,
so let return for now error or warn level for them

mmelnik: Let's make as warnings for now.

PR #376, opened 2020-03-12
EPMLABSBRN-311 remove duplicated test case <eom>

mmelnik:
EPMLABSBRN-311 remove duplicated test case <eom>

### DIFF
--- a/src/test/kotlin/com/epam/brn/controller/UserDetailControllerTest.kt
+++ b/src/test/kotlin/com/epam/brn/controller/UserDetailControllerTest.kt
@@ -76,19 +76,6 @@ internal class UserDetailControllerTest {
         }
 
         @Test
-        fun `should get user by userName`() {
-            // GIVEN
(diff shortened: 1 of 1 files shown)


src/test/kotlin/com/epam/brn/controller/UserDetailControllerTest.kt
```diff
@@ -32,12 +32,12 @@ internal class UserDetailControllerTest {
...
-            lastName = "testUserLastName",
-            email = "unittest@test.ru",
-            active = true,
-            password = "pwd"
+                id = userId,
+                firstName = "testUserFirstName",
+                lastName = "testUserLastName",
+                email = "unittest@test.ru",
+                active = true,
+                password = "pwd"
```
ElenaSpb: it looks strange..
how it relates to "duplicated test case <eom>" ?

ElenaSpb: is it sonar recomendation?

PR #378, opened 2020-03-12
Epmlabsbrn 311 remove non working suppress warning annoations

mmelnik:
Epmlabsbrn 311 remove non working suppress warning annoations <eom>

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/InitialDataLoader.kt
+++ b/src/main/kotlin/com/epam/brn/service/InitialDataLoader.kt
@@ -40,6 +40,8 @@ import org.springframework.core.io.ResourceLoader
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.stereotype.Service
 
+// TODO: this class smells like GOD-object anti-pattern, it should be split
+
 /**
  * This class is responsible for
  * loading seed data on startup.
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/epam/brn/service/InitialDataLoader.kt
```diff
@@ -40,6 +40,8 @@ import org.springframework.core.io.ResourceLoader
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.springframework.stereotype.Service
 
+// TODO: this class smells like GOD-object anti-pattern, it should be split
+
```
ElenaSpb: we have 306 yask for that, can you review it, please

src/test/kotlin/com/epam/brn/service/impl/AuthenticationBasicServiceImplTest.kt
```diff
@@ -57,12 +56,12 @@ internal class AuthenticationBasicServiceImplTest {
     fun `should register new user`() {
         // GIVEN
         val email = "testUser"
-        val password = "testPassword"
+        val passw = "testPassword"
```
ElenaSpb: does it really solve sonar issues?

src/main/kotlin/com/epam/brn/service/InitialDataLoader.kt
mmelnik: So then it will be deleted.

src/test/kotlin/com/epam/brn/service/impl/AuthenticationBasicServiceImplTest.kt
mmelnik: yes

PR #381, opened 2020-03-12
EPMLABSBRN-283 devops: add run tests for angular part of project

DmitriyStoyanov:
[EPMLABSBRN-283](https://jira.epam.com/jira/browse/EPMLABSBRN-283)

**Description**:
devops: add run tests for angular part of project

### DIFF
--- a/.github/workflows/frontend-angular-test.yml
+++ b/.github/workflows/frontend-angular-test.yml
@@ -1,4 +1,4 @@
-name: Frontend tests
+name: Frontend Angular tests
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


.github/workflows/frontend-angular-test.yml
```diff
@@ -0,0 +1,25 @@
+name: Frontend Angular tests
+
+on:
+  pull_request:
+    paths:
+      - 'Dockerfile_frontend'
+      - 'frontend/**'
```
ElenaSpb: why frontend?
maybe it should be frontend-angular?

DmitriyStoyanov: Thank you, for your finding, fixed

PR #452, opened 2020-03-31
EPMLABSBRN-221 rework StudyHistoryService to correctly handle incorrect user/exercise ids <eom>

mmelnik:
[EPMLABSBRN-221](https://jira.epam.com/jira/browse/EPMLABSBRN-221)

500 error was returned because of DB constraint violation. That happened because of missing non-existing dependent entities check, like Exercise and UserAccount. Now if wrong ids are passed and no entities were found correct exception is thrown.

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
+++ b/src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
@@ -6,58 +6,70 @@ import com.epam.brn.exception.EntityNotFoundException
 import com.epam.brn.model.Exercise
 import com.epam.brn.model.StudyHistory
 import com.epam.brn.model.UserAccount
+import com.epam.brn.repo.ExerciseRepository
 import com.epam.brn.repo.StudyHistoryRepository
+import com.epam.brn.repo.UserAccountRepository
 import java.security.InvalidParameterException
-import javax.persistence.EntityManager
+import java.util.Optional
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/epam/brn/service/StudyHistoryService.kt
```diff
@@ -6,58 +6,73 @@ import com.epam.brn.exception.EntityNotFoundException
...
         studyHistoryRepository.save(studyHistory)
-        val studyDto = studyHistory.toDto()
-        studyDto.responseCode = if (isNew) HttpStatus.CREATED else HttpStatus.OK
-        return studyDto
+
+        return studyHistory.toDto()
+    }
+
+    private fun findUserAccount(studyHistoryDto: StudyHistoryDto): UserAccount {
+        val userId = ofNullable(studyHistoryDto.userId)
```
goopnigoop: Looks like we already have preliminary validation to StudyHistoryDto using @Validated to avoid null fields.
I'm also familiar with a rule from Java:
[https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1695](https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1695), but I'm not sure about kotlin and annotation @NotNull fro example.
I'm a little bit confused when i see throwing the NPE explicitly.
@mmelnik what do you think?

PR #683, opened 2020-06-07
Epmlabsbrn-497 [Demo] add multilanguage support for grous

goopnigoop:
[EPMLABSBRN-497](https://jira.epam.com/jira/browse/EPMLABSBRN-497)

**Internalization was implemented for requests to get exercise groups**

For implementation was used a map to store the mapping:
- group name (unique) **to** messages

This solution includes a common spring approach, but the other part is temporary and to implement full internalization flow, translate of elements should be added initially by an operator and saved to the database or translate should be used from messages.properties

Examples:

- api/groups?locale=en_EN

image

- api/groups?locale=ru_RU

image

### DIFF
--- a/src/main/kotlin/com/epam/brn/config/WebMvcBasicConfiguration.kt
+++ b/src/main/kotlin/com/epam/brn/config/WebMvcBasicConfiguration.kt
@@ -1,0 +1,25 @@
+package com.epam.brn.config
+
+import java.util.Locale
+import org.springframework.context.annotation.Bean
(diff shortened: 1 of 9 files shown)


lifeart:
It's should be new records, for english version, instead of translated

goopnigoop:
> It's should be new records, for english version, instead of translated

It requires a discussion, @mmelnik @ElenaSpb let's schedule a meeting after the demo

PR #692, opened 2020-06-09
feature: localization for all main models

lifeart:


ElenaSpb:
adding additional columns in input files need to make additional work in processing files, so only after that it would work correctly

ElenaSpb:
this all would be done in separate task

PR #745, opened 2020-06-25
Fix:  [Authorization] Non-userfriendly error is shown when login has no '@'

mike050210:
[EPMLABSBRN-340](https://jira.epam.com/jira/browse/EPMLABSBRN-340):  [Authorization] Non-userfriendly error is shown when login has no '@'

**Description**:
These changes set pre-defined Key messages to DTO data classes for return specific field error messages in the HTTP response.

Front-end requires specific non-language-dependent messages to handle them in an easier way.


src/main/kotlin/com/epam/brn/config/WebMvcBasicConfiguration.kt
```diff
@@ -1,22 +1,40 @@
...
 
 @Configuration
 class WebMvcBasicConfiguration : WebMvcConfigurer {
 
     @Bean
     fun localeResolver(): LocaleResolver {
-        val slr = SessionLocaleResolver()
-        slr.setDefaultLocale(Locale("ru", "RU"))
-        return slr
+        val localeResolver = AcceptHeaderLocaleResolver()
```
goopnigoop, line 20: If we change to AcceptHeaderLocaleResolver we have to notify our frontend team that we're going to accept a locale value as a part of header _**accept-language**_
@lifeart please, be aware for the future

lifeart: I don't think we need to support header-based localization in API, prefer explicit param in requests

src/main/resources/messages_en.properties
```diff
@@ -1,4 +1,6 @@
 group.first.name=non-speech exercises
 group.second.name=speech exercises
 group.first.description=non-speech exercises
-group.second.description=speech exercises
\ No newline at end of file
+group.second.description=speech exercises
+group.validation.message.email=E-mail must contain '@'
```
lifeart, line 5: we don't need to double localization in be&fe, I prefer to have constaint words on BE and have only one point of system message localization (on FE)

src/main/kotlin/com/epam/brn/config/WebMvcBasicConfiguration.kt
```diff
@@ -1,22 +1,40 @@
...
+        localeResolver.defaultLocale = Locale("ru", "RU")
+        return localeResolver
+    }
+
+    @Bean
+    fun messageSource(): MessageSource {
+        val messageSource = ReloadableResourceBundleMessageSource()
+        messageSource.setBasename("classpath:messages")
+        messageSource.setDefaultEncoding("UTF-8")
```
ElenaSpb, line 29: I suppose there is const for this encoding
like
StandardCharsets.UTF_8.name()

src/main/resources/messages_en.properties
mike050210: Which language would you prefer from the BE?

ElenaSpb: Russia, I suppose

src/main/kotlin/com/epam/brn/config/WebMvcBasicConfiguration.kt
mike050210: Resolved

src/main/resources/messages_en.properties
mike050210: As per a conversation in PR #748, errors would be handle in Front-End side. So the BE is required to return specific Key Errors in standard format.

With the latest changes it would return something like:
```
{
    "data": [],
    "errors": [
        {
            "field": "loginDto:username",
            "error": "INVALID_EMAIL_FORMAT"
        },
        {
            "field": "loginDto:password",
            "error": "EMPTY_FIELD"
        }
    ],
    "meta": []
}
```

ElenaSpb:
great, thanks!

PR #746, opened 2020-06-25
EPMLABSBRN-414 #Registration

SiestaVie:
[EPMLABSBRN-414](https://jira.epam.com/jira/browse/EPMLABSBRN-414)

**Description**:new ExceptionHandler was added and bean- validaator for messageSource


words_series1.txt
```diff
@@ -0,0 +1 @@
+бал,бум,быль,вить,гад,дуб,линь,лис,моль,пар,пять,раб,рак,рожь,сеть,топь,ход,шеф
```
goopnigoop, line 1: Hi!
Just a question why do we have this files as a part of this commit?

PR #763, opened 2020-07-06
EPMLABSBRN-527 Speech generation using javax.speech package

ElenaSpb:
[EPMLABSBRN-527](https://jira.epam.com/jira/browse/EPMLABSBRN-527)

quality is not very good - so it would not be merged. just save there for history a later invistgation


ElenaSpb:
quality is not very good - so it would not be merged. just save there for history a later invistgation

PR #783, opened 2020-07-10
EPMLABSBRN-527 Speech generation using yandex cloud

ElenaSpb:
[EPMLABSBRN-527](https://jira.epam.com/jira/browse/EPMLABSBRN-527) Speech generation using yandex cloud

done


ElenaSpb:
@DmitriyStoyanov please take a look that some creds does not loaded

PR #809, opened 2020-07-16
EPMLABSBRN-363 solving case-sens in email

SiestaVie:
[EPMLABSBRN-363](https://jira.epam.com/jira/browse/EPMLABSBRN-363)

**Description**:


src/main/kotlin/com/epam/brn/dto/LoginDto.kt
```diff
@@ -9,7 +9,7 @@ data class LoginDto(
     val grant_type: String = "password",
     @field:NotBlank(message = "{validation.field.email.blank}")
     @field:Email(message = "{validation.field.email.invalid-format}")
-    val username: String,
+    var username: String,
```
ElenaSpb, line 12: why it become `var`?

SiestaVie: solved

PR #856, opened 2020-08-02
EPMLABSBRN-564 add check on generation audio files step: create only for new words

ElenaSpb:
[EPMLABSBRN-564](https://jira.epam.com/jira/browse/EPMLABSBRN-564)

add check on generation audio files step: create only for new words


src/main/kotlin/com/epam/brn/service/AudioFilesGenerationService.kt
```diff
@@ -174,4 +190,34 @@ class AudioFilesGenerationService(@Autowired val wordsService: WordsService) {
...
+        val request = HttpEntity(null, headers)
+
+        val response: ResponseEntity<ByteArray> = restTemplate
+                .exchange(uriBuilder.build(), HttpMethod.POST, request, ByteArray::class.java)
+        if (response.statusCode != HttpStatus.OK)
+            throw YandexServiceException("Yandex cloud does not provide audio file for word `$word`, httpStatus={${response.statusCode}}")
+        log.info("Ogg audio file for Word `$word` was successfully generated.")
+
+        val byteArray = response.body!!
+        val fileOgg = File("$word.ogg")
```
lifeart, line 219: note, we should not store files with lang-specific symbols (cyrillic or such), let's have `md5` hash as file name.

ElenaSpb: in https://jira.epam.com/jira/browse/EPMLABSBRN-578

PR #918, opened 2020-08-30
EPMLABSBRN-577 Support noise url in config file and in exercises

goopnigoop:
[EPMLABSBRN-577](https://jira.epam.com/jira/browse/EPMLABSBRN-577)

The current task provides the support for noiseLevel and noiseUrl columns for CSV files.

- [x] updated the exercise table to include 2 new columns

- [x] updated parsing logic

- [x] updated unit/integration test


src/main/kotlin/com/epam/brn/upload/csv/series1/SeriesOneRecordProcessor.kt
```diff
@@ -95,7 +95,8 @@ class SeriesOneRecordProcessor(
                     series = series,
                     name = record.exerciseName,
                     level = record.level,
-                    noise = record.noise,
+                    noiseLevel = record.noiseLevel,
+                    noiseUrl = record.noiseUrl,
```
ElenaSpb, line 99: I can not find where is setting this parameter from file to our model to save it to DB. ?

goopnigoop: Here, in extractExercise method, we set it to exercise before saving

ElenaSpb (approved):

PR #1261, opened 2020-12-04
EPMLABSBRN-655. Put endpoint for current user avatar.

asus200786:
[EPMLABSBRN-655](https://jira.epam.com/jira/browse/EPMLABSBRN-655)

**Description**: Created PUT endpoint for update current user avatar.


api-contract/api.raml
```diff
@@ -203,6 +203,39 @@ mediaType:
...
+      get:
+        description: Get a user avatar.
+        responses:
+          200:
+            body:
+              application/json:
+                example: !include samples/avatar_get_response.json
+
+      post:
+        description: Create user avatar.
```
ElenaSpb, line 217: any body here?

api-contract/api.raml
```diff
@@ -203,6 +203,39 @@ mediaType:
...
+                example: !include samples/avatar_get_response.json.json
+
+      put:
+        is: [Forbidden, NotFound]
+        desctiprion: Update user avatar.
+        queryParameters:
+          avatarUrl:
+              required: true
+              type: string
+              example: {schema}://{host}/api/users/current/avatar?avatarUrl=lime.xml
```
ElenaSpb, line 231: example would be like "/pictures/avatar/girl.cvg" as I understand

api-contract/api.raml
ElenaSpb: and I did not find realization

src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
```diff
@@ -35,4 +36,11 @@ class UserDetailController(@Autowired val userAccountService: UserAccountService
         @RequestParam("name", required = true) name: String
     ) = ResponseEntity.ok()
         .body(BaseResponseDto(data = listOf(userAccountService.findUserByName(name))))
+
+    @PutMapping(value = ["/current/avatar"])
```
ElenaSpb, line 40: could you please add integration test for this new end-point

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -91,6 +91,13 @@ class UserAccountServiceImpl(
     override fun getUsers(): List<UserAccountResponse> =
         userAccountRepository.findAll().map { it.toDto() }
 
+    override fun updateAvatarCurrentUser(avatarUrl: String): UserAccountResponse {
```
ElenaSpb, line 94: could you please add unit test for thus new functionallity

PR #1318, opened 2020-12-16
EPMLABSBRN-622: csv processor for non speech exercises

vika95viktoria:
[EPMLABSBRN-622](https://jira.epam.com/jira/browse/EPMLABSBRN-622)

**Description**:
Added CSV file and corresponding handlers for non speech exercises
Added encoding setting for console logger to display Russian words correctly


src/main/kotlin/com/epam/brn/service/InitialDataLoader.kt
```diff
@@ -70,7 +67,8 @@ class InitialDataLoader(
         fileNameForSeries(1),
         fileNameForSeries(2),
         fileNameForSeries(3),
-        fileNameForSeries(4)
+        fileNameForSeries(4),
+        "non_speech.csv"
```
ElenaSpb, line 71: as we plan to add series into nonspeech group 
it would be reasonable to add config file for each of them.
so lets name it like in first group 

fileNameForSeries(11)
fileNameForSeries(12)

ElenaSpb: as it is now structurally here https://docs.google.com/document/d/1eyrumCWOv2AU50NoFlT3vZQ6RSbIgHxLdekj-FFD1hM/edit#heading=h.5u2lpqs06eeo

vika95viktoria: Current implementation incorporates series into csv structure. Wouldn't it be more useful to have everything in config file instead of manually fetching series by magical numbers(id=11, etc)?

src/main/kotlin/com/epam/brn/upload/csv/nonspeech/NonSpeechRecord.kt
```diff
@@ -0,0 +1,21 @@
+package com.epam.brn.upload.csv.nonspeech
+
+import com.epam.brn.model.ExerciseType
+import com.fasterxml.jackson.annotation.JsonProperty
+
+data class NonSpeechRecord(
+    @JsonProperty("series")
```
asundukov, line 7: а здесь не надо @field: ?

vika95viktoria: @asundukov not sure that I understand you correctly, could you, please, expand a little bit on your idea?

asundukov: yeah ok it is just about annotate field, not constructor parameter, but in this case it seems that it doesnt matter

ElenaSpb (approved):

PR #1423, opened 2021-01-23
EPMLABSBRN-681 update Kotlin version, use coroutins

ElenaSpb:
[EPMLABSBRN-681](https://jira.epam.com/jira/browse/EPMLABSBRN-681)

update Kotlin version, use coroutins


ElenaSpb:
done in separate

PR #1500, opened 2021-02-10
EPMLABSBRN-698 add rest api for audiometry

ElenaSpb:
[EPMLABSBRN-698](https://jira.epam.com/jira/browse/EPMLABSBRN-698)

add rest api for audiometry


api-contract/api.raml
```diff
@@ -6,6 +6,29 @@ baseTestUri: http://10.66.216.143:8081/api
 mediaType:
   - application/json
 
+# AUDIOMETRY
+/audiometrics
+    get:
+        queryParameters:
+            locale:
```
lifeart, line 13: why we need locale here? locale

api-contract/api.raml
```diff
@@ -6,6 +6,29 @@ baseTestUri: http://10.66.216.143:8081/api
...
+                required: false
+                type: String
+                example: "ru-ru"
+                default: "ru-ru"
+        responses:
+            200:
+                body:
+                    example: !include samples/audiometrics.json
+
+/audiometryHistory
```
lifeart, line 23: let's don't use camelCase in endpoint names,
`audiometryHistory` -> `audiometry-history`, 

anyway I'm not celarly understand meaning of this API

api-contract/api.raml
ElenaSpb: yes, as we have audiometryType=SPEECH

ElenaSpb: in this case content would be provide by locale

api-contract/api.raml
ElenaSpb: to save speech diagnostic results

ElenaSpb: done: audiometryHistory -> audiometry-history

PR #1508, opened 2021-02-12
EPMLABSBRN-763 BE: implement rest api for getting done exercises in subgroup

nikolai4lazarev:
EPMLABSBRN-763 https://jira.epam.com/jira/browse/EPMLABSBRN-763#

New end-point added to get user progress (done/all) for subGroups.
Unit/Integration tests. 
api contract extended by information about new end-point


src/main/kotlin/com/epam/brn/controller/SubGroupController.kt
```diff
@@ -19,7 +19,12 @@ class SubGroupController(val subGroupsService: SubGroupService) {
 
     @GetMapping
     @ApiOperation("Get subGroups for series id.")
-    fun getAllGroups(@RequestParam(value = "seriesId", required = true) seriesId: Long): ResponseEntity<BaseResponseDto> {
+    fun getAllGroups(
+        @RequestParam(
+            value = "seriesId",
+            required = true
+        ) seriesId: Long
+    ): ResponseEntity<BaseResponseDto> {
```
ElenaSpb, line 27: there were no changes here, please, return old format to remove this changes from MR

ElenaSpb (approved):

api-contract/api.raml
```diff
@@ -89,6 +89,29 @@ mediaType:
...
+          description: "List of IDs of subGroups"
+          required: true
+          type: List<Long>
+          example: "1,2,3"
+      responses:
+        200:
+          body:
+            example: {
+              data: {
+                "1": {
```
lifeart, line 105: I think we have to implement preagreggation on backend side, this data should be used for plots?

cc @ElenaSpb

ElenaSpb: @lifeart return should be map:  subGroupId: Pair<done/avaliable>
yes?

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
```diff
@@ -28,6 +35,17 @@ class SubGroupService(private val subGroupRepository: SubGroupRepository) {
             .orElseThrow { EntityNotFoundException("No subGroup was found by id=$subGroupId.") }
         return subGroup.toDto(pictureTheme)
     }
+
+    fun getSubGroupsProgressForUser(subGroupIds: List<Long>): Map<Long, Map<Int, Int>> {
+        log.debug("Trying to get done/all exercises for each subGroup. subGroupIds=$subGroupIds")
```
ElenaSpb, line 40: Map<Long, Map<Int, Int>>
lets change on
Map<Long, Pair<Int, Int>>

api-contract/api.raml
ElenaSpb: with 
return subGroupIds.map {
            it to Pair(
                studyHistoryRepository.getDoneExercises(it, currentUser.id!!).size,
                exerciseRepository.findExercisesBySubGroupId(it).size
            )
        }.toMap()

ElenaSpb: image
?

ElenaSpb: or you want something like 
image

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
nikolai4lazarev: There is one problem with Pair. As I remember, if I will send Map<Long, Pair<Int, Int>> then front-end will get JSON like 
`1: {
"first": 1,
"second": 2
}`
I'm not sure that it will be comfortable for front-end have a deal with "first" and "second"

api-contract/api.raml
```diff
@@ -89,6 +89,33 @@ mediaType:
         200:
           body:
             example: !include samples/subgroup_by_id.json
+  /statistic:
```
lifeart, line 92: let's have less generic name for this query, like /statistics/subgroups

ElenaSpb (approved):

lifeart (approved):

PR #1830, opened 2021-04-09
EPMLABSBRN-835: BE: add Description into Resource model and support it…

richardforgo:
[EPMLABSBRN-835](https://jira.epam.com/jira/browse/EPMLABSBRN-835) BE: add Description into Resource model and support it in appl

**Description**:
- Added Description field for Resource
- The field is now supported in reads (added to ResourceDto)
- Added new admin endpoint POST admin/resources/{id}?description=whatever-description
for updating the description for existing Resources.
- Updated unit and integration tests


api-contract/api.raml
```diff
@@ -441,6 +441,17 @@ mediaType:
             responses:
                   200:
                     example: !include samples/study_histories_dto_response.json
+    /resources:
+        /{id}:
+          post:
```
goopnigoop, line 446: Hi!. Why POST is used for an update operation, but not a PUT or PATCH?

richardforgo: Hey @goopnigoop, good point.
I've updated this to PATCH instead in https://github.com/Brain-up/brn/pull/1830/commits/27e1364bf64238c81ba29e3efc8df8b905b62ffa

richardforgo:
https://github.com/Brain-up/brn/pull/1830/commits/27e1364bf64238c81ba29e3efc8df8b905b62ffa
-Modified the new endpoint to be PATCH instead of POST.
-Deleted 404 case from Integration tests (should be used for happy path only afaik)
-updated all tests with given/when/then comments

src/main/resources/db/migration/V04092021_835.sql
```diff
@@ -0,0 +1 @@
+alter table resource add column description varchar(255);
```
richardforgo, line 1: please take a look at the versioning. I've used the date and the ticket number for this.

ElenaSpb: it should be V220210420_835
where 20210420 - current date

richardforgo: updated

ElenaSpb (approved):

PR #1935, opened 2021-05-07
[EPMLABSBRN-850](https://jira.epam.com/jira/browse/EPMLABSBRN-850)

relizerel:
[EPMLABSBRN-850](https://jira.epam.com/jira/browse/EPMLABSBRN-850)

**Description**:
Created the role file and added the required fields to the entity and DTO


src/main/kotlin/com/epam/brn/model/Role.kt
```diff
@@ -0,0 +1,7 @@
+package com.epam.brn.model
+
+enum class Role(s: String) {
```
ElenaSpb, line 3: please change 
s -> name

richardforgo (approved):
Looks good to me, but shouldn't there be some database changes?

ElenaSpb:
in another MR will be done

PR #1954, opened 2021-05-10
[EPMLABSBRN-841] BE: change user data model: add photo, doctor, description

richardforgo:
[EPMLABSBRN-841](https://jira.epam.com/jira/browse/EPMLABSBRN-841)

**Description**:
Modified UserAccount data model:
- added 1-* relationship for doctors to add patients
- added foto for users to upload a profile picture
- added description


src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -130,6 +130,14 @@ class UserAccountServiceImpl(
             currentUserAccount.avatar = userChangeRequest.avatar
             changed = true
         }
+        if (userChangeRequest.picture != null) {
```
nikolai4lazarev, line 133: All these IFs look really scary. If in the future we will have 10 or even 20 fields for the user then we will have 20 IFs. Furthermore, if we need to add a new field for the user then we need to change this code, and that's bad (it violates the Open-closed Principle). There is my suggestion, let's use a spring-based "chain of responsibility" pattern here and make the refactoring now while we don't have 20 IFs, it will make our code Open for extending and close for changing. Please, think about it. If you don't know what "chain of responsibility" is, then you can find a lot of examples out of here on the internet

src/main/kotlin/com/epam/brn/dto/request/UserAccountCreateRequest.kt
```diff
@@ -31,7 +32,10 @@ data class UserAccountCreateRequest(
     val bornYear: Int,
     @field:NotNull(message = "{validation.field.gender.notNull}")
     val gender: Gender,
-    val avatar: String? = null
+    val avatar: String? = null,
+    val foto: String? = null,
```
goopnigoop, line 36: I'd like to propose to rename **foto**->**photo** just to follow the rules

src/main/kotlin/com/epam/brn/model/UserAccount.kt
```diff
@@ -31,7 +32,12 @@ data class UserAccount(
     var created: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")),
     @Column(nullable = false)
     var changed: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")),
-    var avatar: String? = null
+    var avatar: String? = null,
```
ElenaSpb, line 35: please do not forget about flyway scripts,
to make our already run application workable.
you need to add scripts db.migration folder

src/main/kotlin/com/epam/brn/dto/request/UserAccountChangeRequest.kt
```diff
@@ -10,5 +11,8 @@ data class UserAccountChangeRequest(
     @field:Min(value = 1900, message = "{validation.field.bornYear.moreThen1900}")
     val bornYear: Int? = null,
     val gender: Gender? = null,
-    val avatar: String? = null
+    val avatar: String? = null,
+    val picture: String? = null,
```
ElenaSpb, line 15: photo?

src/main/kotlin/com/epam/brn/model/UserAccount.kt
```diff
@@ -31,7 +32,12 @@ data class UserAccount(
...
+    var foto: String? = null,
```
ElenaSpb, line 36: also is Strin a suitable type to save picture file ?
lets think about it carefully

richardforgo: Yeah, I was thinking it would be an URL to a picture. Would we instead store the binary in the database as a blob?

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
richardforgo: updated this part

src/main/kotlin/com/epam/brn/model/UserAccount.kt
richardforgo: Tested them, I've added "if not exists" to "add column" statements

src/main/kotlin/com/epam/brn/dto/request/UserAccountChangeRequest.kt
richardforgo: updated

src/main/kotlin/com/epam/brn/model/UserAccount.kt
richardforgo: @ElenaSpb let me know if I should change this

nikolai4lazarev (approved):

src/main/kotlin/com/epam/brn/dto/request/UserAccountChangeRequest.kt
```diff
@@ -10,5 +11,8 @@ data class UserAccountChangeRequest(
     @field:Min(value = 1900, message = "{validation.field.bornYear.moreThen1900}")
     val bornYear: Int? = null,
     val gender: Gender? = null,
-    val avatar: String? = null
+    val avatar: String? = null,
+    val photo: String? = null,
+    @field:Length(min = 1, max = 255)
```
ElenaSpb, line 16: why max = 255?
I think it may be not enough for description

src/main/resources/db/migration/V220210605_841.sql
```diff
@@ -0,0 +1,7 @@
+alter table user_account add column if not exists description varchar(255);
+alter table user_account add column if not exists photo varchar(255);
+alter table user_account
+        add column if not exists doctor bigint,
+        add constraint a47569ad8b6996ef5d8eea2e73d1e2ab
```
ElenaSpb, line 5: maybe the name of constraint may be readable?
user_account_doctor_constraint

src/main/resources/db/migration/V220210605_841.sql
```diff
@@ -0,0 +1,7 @@
+alter table user_account add column if not exists description varchar(255);
+alter table user_account add column if not exists photo varchar(255);
```
ElenaSpb, line 2: please delete the limitation of 255

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -138,31 +138,32 @@ class UserAccountServiceImpl(
...
-            currentUserAccount.toDto()
+        return getCurrentUser().let {
+            if (userChangeRequest.isNotEmpty())
+                userAccountRepository.save(it.updateFields(changeRequest = userChangeRequest))
+            else it
+        }.toDto()
     }
 
+    private fun UserAccountChangeRequest.isNotEmpty(): Boolean =
+        this.name.isNullOrBlank()
```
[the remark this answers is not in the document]
richardforgo: you're right, I've fixed this. Good catch!

src/main/resources/db/migration/V220210420_835.sql
```diff
@@ -1 +1 @@
-alter table resource add column description varchar(255);
\ No newline at end of file
+alter table resource add column if not exists description varchar(255);
```
ElenaSpb, line 1: could you please delete limitation here too. ?

ElenaSpb (approved):

src/main/kotlin/com/epam/brn/model/UserAccount.kt
```diff
@@ -34,6 +34,11 @@ data class UserAccount(
     @Column(nullable = false)
     var changed: ZonedDateTime = ZonedDateTime.now(ZoneId.of("UTC")),
     var avatar: String? = null,
+    var photo: String? = null,
+    var description: String? = null,
+    @OneToMany
+    @JoinColumn(name = "doctor", referencedColumnName = "id")
+    var patients: MutableList<UserAccount> = mutableListOf(),
```
ElenaSpb, line 41: as you add column `doctor` into userAccount model in script - maybe it should be added here too?

ElenaSpb (approved):

PR #1985, opened 2021-05-18
Issue 1971: Unit tests for AdminController, SubGroupService, SubGroupController

WikkiRin:
ISSUE-1971: BE Unit tests 3

Unit tests for AdminController, SubGroupService, SubGroupController


build.gradle
```diff
@@ -57,6 +57,7 @@ dependencies {
         exclude group: "org.mockito"
     }
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
+    testImplementation "io.mockk:mockk:$mockkVersion"
```
nikolai4lazarev, line 60: @ElenaSpb what do u think about using MockK here? Veronica decided to use it because of tests of extension functions. We researched it together and came to the conclusion that we can not mock an extension function using Mockito

src/test/kotlin/com/epam/brn/controller/AdminControllerTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun `getUsers should return users with statistic when withAnalytics is true`() {
+        // GIVEN
+        val withAnalytics = true
+        `when`(userAccountService.getUsersWithAnalytics(pageable)).thenReturn(listOf(userWithAnalyticsDto))
+
+        // WHEN
+        val users = adminController.getUsers(withAnalytics, pageable)
+
+        // THEN
+        verify(userAccountService, times(1)).getUsersWithAnalytics(pageable)
```
ElenaSpb, line 85: times(1) is by default, you can miss it in check

build.gradle
ElenaSpb: yes, we think about using mockk for several month, maybe we can start to use it
and maybe switch on it in exist tests.?

src/test/kotlin/com/epam/brn/service/SubGroupServiceTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        val subGroupDto = SubGroupDto(seriesId, 1L, 5, "name", "pictureURL", "description")
+        mockkStatic("com.epam.brn.service.SubGroupServiceKt")
+        every { subGroupRepository.findBySeriesId(seriesId) } returns listOf(subGroup)
+        every { subGroup.toDto(pictureTheme) } returns subGroupDto
+
+        // WHEN
+        val allGroups = subGroupService.findSubGroupsForSeries(seriesId)
+
+        // THEN
+        verify { subGroupRepository.findBySeriesId(seriesId) }
```
koylubaevnt, line 51: I think need change to verify(exactly = 1), because by default MockK use atLeast=1 (1  and more times - is passed) and this is not equal to Mockito verify().
If we must check exactly call times.

build.gradle
nikolai4lazarev: It will take time if we talk about all tests. If only about these new tests then it's ok to do it in this task

goopnigoop: I agree with @nikolai4lazarev I think that's a good idea to use it for new tests and do not waste a time on the old ones

ElenaSpb: as we have team of 3 junior developers now,
maybe it would be reasonable for them take couple tests for refactoring.
?

nikolai4lazarev: Think the idea is good. But if we want to use MockK now, then we should let know to juniors' team as soon as possible, because they write tests right now using Mockito

nikolai4lazarev: Actually, we should let know to all other back-end developers too

nikolai4lazarev: @ElenaSpb what do u think?

nikolai4lazarev (approved):

Issue #2005, opened 2021-05-31
BE unit tests for AuthorityServiceImpl, CloudController


nikolai4lazarev:
Write unit tests for 
AuthorityServiceImpl,
CloudController

Use MockK in your test

PR #2011, opened 2021-06-01
Unit tests for AuthorityServiceImpl, CloudController

rkonoplev:
[ISSUE-2005](https://github.com/Brain-up/brn/issues/2005)

**Description**:
Unit tests for AuthorityServiceImpl, CloudController


src/test/kotlin/com/epam/brn/auth/AuthorityServiceImplTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+    @Test
+    fun `should save authority`() {
+        // GIVEN
+        val authority = Authority(
+            id = 1L,
+            authorityName = "FirstName"
+        )
+        val authoritySaved = Authority(
```
nikolai4lazarev, line 96: There is no reason to have a copy of your authority. You can just return the authority from the save method

src/test/kotlin/com/epam/brn/auth/AuthorityServiceImplTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    @Test
+    fun `should throw error when authority is null`() {
+        // GIVEN
+        val authority = Authority(
+            id = 1L,
+            authorityName = "FirstName"
+        )
+        every { authorityRepository.save(authority) } throws IllegalArgumentException("Authority is not found")
+        // WHEN
+        assertFailsWith<IllegalArgumentException> {
```
nikolai4lazarev, line 116: You should test the save method from the service, not from the repository

src/test/kotlin/com/epam/brn/auth/AuthorityServiceImplTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        assertEquals(authority, resultSaving)
+    }
+    @Test
+    fun `should throw error when authority is null`() {
+        // GIVEN
+        val authority = Authority(
+            id = 1L,
+            authorityName = "FirstName"
+        )
+        every { authorityRepository.save(authority) } throws IllegalArgumentException("Authority is not found")
```
nikolai4lazarev, line 114: This message is not really clear, because u actually try to save an object but not find it. U can just throw an exception without the message.
Also, it seems a little bit strange, that your method has "when authority is null" in the name, but you actually pass an authority that is not null

src/test/kotlin/com/epam/brn/auth/AuthorityServiceImplTest.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    @Test
+    fun `should throw error when authority by Name is not found`() {
+        // GIVEN
+        val authorityName = "Name"
+        every { authorityRepository.findAuthorityByAuthorityName(authorityName) } throws EntityNotFoundException("Authority is not found")
+
+        // WHEN
+        assertFailsWith<EntityNotFoundException> {
+            authorityRepository.findAuthorityByAuthorityName(authorityName)
```
nikolai4lazarev, line 86: Same problem here, u test a repository but you should test a service

rkonoplev: Done

src/test/kotlin/com/epam/brn/auth/AuthorityServiceImplTest.kt
```diff
@@ -0,0 +1,133 @@
...
+        every { authorityRepository.save(authority) } returns (authority)
+        // WHEN
+        val resultSaving = authorityServiceImpl.save(authority)
+        // THEN
+        verify(exactly = 1) { authorityRepository.save(authority) }
+        assertEquals(authority, resultSaving)
+    }
+
+    @Test
+    fun `should throw error when authority is not found`() {
```
nikolai4lazarev, line 105: Seems like this test is usless. U can remove it

nikolai4lazarev (approved):

PR #2015, opened 2021-06-08
EPMLABSBRN-878: Backend: study history endpoint failing with 500

koylubaevnt:
[EPMLABSBRN-878](https://jira.epam.com/jira/browse/EPMLABSBRN-878)

**Description**:
EPMLABSBRN-878: Backend: study history endpoint failing with 500 without error message

* add migration script which: compare maximum id in table with last value of sequence for this table and if table id is bigger or equal to sequence change current value of sequence. It's done for tables: 'signal', 'signal', 'task', 'exercise', 'resource', 'sub_group', 'series', 'study_history'.
* move spring.jpa.show-sql=true from application.properties to application-dev.properties


src/main/resources/db/migration/V220210608_878.sql
```diff
@@ -0,0 +1,19 @@
+DO $$
+DECLARE
+    r record;
+    table_max_value bigint;
+    sequence_cur_value bigint;
+    sequence_name varchar;
+    sequence_postfix varchar = '_id_seq';
+BEGIN
+    FOR r IN SELECT table_name FROM information_schema.tables
+             WHERE table_name in ('signal', 'signal', 'task', 'exercise', 'resource', 'sub_group', 'series', 'study_history')
```
nikolai4lazarev, line 10: Signal for two times?

koylubaevnt: Yep. It's some mistake (duplicates).
But result will be only 7 tables (not 8): because used IN. 
I make correction.
Thank you

nikolai4lazarev (approved):

DmitriyStoyanov (approved):

Dmitriy1Fokin (approved):

Issue #2023, opened 2021-06-16
[frontend] Description & locale buttons are transparent if user is not logged in.


korneev-andrew:
Since they are white-text-colored on a white bg...

**Actual**:
![Screenshot from 2021-06-16 23-56-33](https://user-images.githubusercontent.com/16151917/122294488-c52de100-cf00-11eb-8bb2-bb64a713c1e1.png)

**Expected**
![Screenshot from 2021-06-17 00-12-37](https://user-images.githubusercontent.com/16151917/122294491-c5c67780-cf00-11eb-8ec9-1484e313ae63.png)

Keep in mind that on hover text and bg colors should be swapped (white bg -> purple text, purple bg -> white text)

P.S.: It turnes out that active locale is not bold => fix this too

korneev-andrew:
resolved in #2026

PR #2026, opened 2021-06-17
[frontend] 2023: Description & locale buttons are transparent if user is not logged in

korneev-andrew:
+ getter activeLocale added into the header component

#2023


lifeart (approved):

lifeart:
I see one test failure 
image

we should not have locale query param here

lifeart:
ah, let it be here, but we have to fixt test to not fail

Issue #2034, opened 2021-06-30
[BE] Design & implement rest api for delete subgroup


rkonoplev:
delete end point only for subgroups without linked exercises yet

PR #2041, opened 2021-07-06
Design & implement rest api for delete subgroup

rkonoplev:
[EPMLABSBRN-886](https://github.com/Brain-up/brn/issues/2034) 

**Description**:

delete end point only for subgroups without linked exercises yet

**Details**:

Refactoring and tests for ExerciseRepository, SubGroupController and SubGroupService, ExerciseGroupsService and Integration tests for SubGroupControllerIT,  api description in  raml.api


src/main/kotlin/com/epam/brn/controller/SubGroupController.kt
```diff
@@ -30,4 +31,11 @@ class SubGroupController(private val subGroupsService: SubGroupService) {
         val subGroupDto = subGroupsService.findById(subGroupId)
         return ResponseEntity.ok(BaseSingleObjectResponseDto(data = subGroupDto))
     }
+
+    @DeleteMapping("{subGroupId}")
+    @ApiOperation("Get subGroup for id.")
```
ElenaSpb, line 36: Get subGroup for id ->
Delete subGroup by id

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
```diff
@@ -28,6 +32,14 @@ class SubGroupService(private val subGroupRepository: SubGroupRepository) {
             .orElseThrow { EntityNotFoundException("No subGroup was found by id=$subGroupId.") }
         return subGroup.toDto(pictureTheme)
     }
+
+    fun deleteById(subGroupId: Long): Boolean {
+        if (exerciseRepository.findExercisesBySubGroupId(subGroupId).isNotEmpty()) {
+            throw IllegalArgumentException("Impossible delete. SubGroup is not empty.")
```
nikolai4lazarev, line 38: The message is not really clear. Let's write something like "Can not delete subgroup because there are exercises that refer to the subgroup" for example

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
```diff
@@ -28,6 +32,14 @@ class SubGroupService(private val subGroupRepository: SubGroupRepository) {
             .orElseThrow { EntityNotFoundException("No subGroup was found by id=$subGroupId.") }
         return subGroup.toDto(pictureTheme)
     }
+
+    fun deleteById(subGroupId: Long): Boolean {
```
nikolai4lazarev, line 36: Is there any reason to return a boolean value here? Because if we can not delete the subgroup, we throw an exception but do not return false. In another case, if we did not get the error then it means that we successfully deleted the subgroup, and there is no reason to return true. @ElenaSpb your thoughts?

ElenaSpb (changes requested):
Also please add unit , integration tests,
and api description into raml.api

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
ElenaSpb: let it be void

src/main/kotlin/com/epam/brn/controller/SubGroupController.kt
rkonoplev: Done

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
rkonoplev: Done

src/main/kotlin/com/epam/brn/service/SubGroupService.kt
```diff
@@ -28,6 +32,13 @@ class SubGroupService(private val subGroupRepository: SubGroupRepository) {
             .orElseThrow { EntityNotFoundException("No subGroup was found by id=$subGroupId.") }
         return subGroup.toDto(pictureTheme)
     }
+
+    fun deleteSubGroupById(subGroupId: Long) {
+        if (exerciseRepository.findExercisesBySubGroupId(subGroupId).isNotEmpty()) {
```
Dmitriy1Fokin, line 37: maybe it's better to make a query to db that will return just true/false, than to return a collection of exercises

Dmitriy1Fokin:
if the subgroup does not exist, 500 error is returned.  It may be better to add a presence check to the database. And if there is no subgroup do nothing

Dmitriy1Fokin (changes requested):
if the subgroup does not exist, 500 error is returned. It may be better to add a presence check to the database. And if there is no subgroup do nothing

src/test/kotlin/com/epam/brn/service/SubGroupServiceTest.kt
```diff
@@ -54,7 +59,49 @@ internal class SubGroupServiceTest {
...
+        mockkStatic("com.epam.brn.service.SubGroupServiceKt")
+        every { subGroupRepository.existsById(subGroupId) } returns true
+        every { exerciseRepository.existsBySubGroupId(subGroupId) } returns false
+        justRun { subGroupRepository.deleteById(subGroupId) }
+
+        // WHEN
+        subGroupService.deleteSubGroupById(subGroupId)
+
+        // THEN
+        verify(exactly = 1) { exerciseRepository.findExercisesBySubGroupId(subGroupId) }
```
Dmitriy1Fokin, line 78: this test will fail. Please change findExercisesBySubGroupId to existsBySubGroupId

Dmitriy1Fokin (changes requested):

api-contract/api.raml
```diff
@@ -265,7 +265,25 @@ mediaType:
...
+      queryParameters:
+        subGroupId:
+          description: "ID of the subGroup tasks belongs to"
+          required: true
+          type: string
+          example: "3"
+      responses:
+        204:
+        400:
+        description: |
```
Dmitriy1Fokin, line 280: I think there is not enough indentation here

api-contract/api.raml
```diff
@@ -265,7 +265,25 @@ mediaType:
...
+             Bad or not empty subGroup. This can happen if the user inserts
+             incorrect number of subGroup or subGroup has exercises.
+         body:
+            application/json:
+              type: Error
+              example: { "errors": "Can not delete subGroup because subGroup is not found by this id." }
```
Dmitriy1Fokin, line 286: I think the response will look like this:
{
    "data": [],
    "errors": ["Can not delete subGroup because subGroup is not found by this id."],
    "meta": []
}

api-contract/api.raml
```diff
@@ -265,7 +265,25 @@ mediaType:
...
+          required: true
+          type: string
+          example: "3"
+      responses:
+        204:
+        400:
+        description: |
+             Bad or not empty subGroup. This can happen if the user inserts
+             incorrect number of subGroup or subGroup has exercises.
+         body:
```
Dmitriy1Fokin, line 283: and here

Dmitriy1Fokin (approved):

api-contract/api.raml
```diff
@@ -265,7 +265,29 @@ mediaType:
...
+      description: |
+      The endpoint should delete subGroup without exercises
+      queryParameters:
+        subGroupId:
+          description: "ID of the subGroup tasks belongs to"
+          required: true
+          type: string
+          example: "3"
+      responses:
+        204:
```
ElenaSpb, line 278: maybe better simpler 200 response?

src/main/kotlin/com/epam/brn/service/ExerciseGroupsService.kt
```diff
@@ -18,7 +18,7 @@ class ExerciseGroupsService(
     fun findAllGroups(): List<ExerciseGroupDto> {
         log.debug("Searching all groups")
         val groups: List<ExerciseGroup> = exerciseGroupRepository.findAll()
-        return groups.mapNotNull { group -> group.toDto() }
+        return groups.map { group -> group.toDto() }
     }
```
ElenaSpb, line 22: what the reason of this change?

src/main/kotlin/com/epam/brn/service/ExerciseGroupsService.kt
```diff
@@ -18,7 +18,7 @@ class ExerciseGroupsService(
     fun findAllGroups(): List<ExerciseGroupDto> {
         log.debug("Searching all groups")
         val groups: List<ExerciseGroup> = exerciseGroupRepository.findAll()
-        return groups.mapNotNull { group -> group.toDto() }
+        return groups.map { group -> group.toDto() }
```
ElenaSpb, line 21: what the reason of this change?

rkonoplev: IDEA advised this change and Nikolay approved refactoring.

api-contract/api.raml
rkonoplev: I’ve compared it earlier but the better solution is 204.

ElenaSpb (approved):

ElenaSpb (approved):

PR #2061, opened 2021-08-03
Feature/epmlabsbrn 843 updating rest api for admin

KristevSM:
[EPMLABSBRN-843](https://jira.epam.com/jira/browse/EPMLABSBRN-843)

**Description**: designed and implemented REST API for admin: get roles, users by role


api-contract/api.raml
```diff
@@ -818,6 +822,18 @@ mediaType:
                 body:
                   application/json:
                     example: !include samples/add_subgroup_to_series_response.json
+    /roles:
+                get:
```
ElenaSpb, line 826: we need end-point to get all roles, not by user, just list of the roles

src/test/kotlin/com/epam/brn/controller/AdminControllerTest.kt
```diff
@@ -113,13 +114,14 @@ internal class AdminControllerTest {
...
+        val role = "ROLE_USER"
+        every { userAccountService.getUsers(pageable, role) } returns listOf(userAccountDto)
 
         // WHEN
-        val users = adminController.getUsers(withAnalytics, pageable)
+        val users = adminController.getUsers(withAnalytics, role, pageable)
 
         // THEN
-        verify(exactly = 1) { userAccountService.getUsers(pageable) }
+        verify(exactly = 1) { userAccountService.getUsers(pageable, role) }
```
ElenaSpb, line 124: could you please add integration test to test get users by role

src/main/kotlin/com/epam/brn/repo/UserAccountRepository.kt
```diff
@@ -23,4 +23,7 @@ interface UserAccountRepository : JpaRepository<UserAccount, Long> {
 
     @Query("select DISTINCT u FROM UserAccount u left JOIN FETCH u.authoritySet left JOIN FETCH u.headphones where u.doctor.id = ?1")
     fun findUserAccountsByDoctorId(doctorId: Long): List<UserAccount>
+
+    @Query("select DISTINCT u FROM UserAccount u left JOIN FETCH u.authoritySet authorities left JOIN FETCH u.headphones where authorities.authorityName = ?1")
```
Dmitriy1Fokin, line 27: why is "left JOIN FETCH u.headphones" here?. And it is better to use named parameters instead of position-based parameter

KristevSM: I used the findUsersAccountsByRole method instead of findAll(), so I get the full UserAccount object

ElenaSpb (approved):

PR #2069, opened 2021-08-06
EPMLABSBRN-899 change audiometry dto, fix loading sin audiometry

ElenaSpb:
[EPMLABSBRN-899](https://jira.epam.com/jira/browse/EPMLABSBRN-899)

change dto, fix loading sin audiometry


src/main/kotlin/com/epam/brn/service/load/AudiometryLoader.kt
```diff
@@ -92,6 +92,11 @@ class AudiometryLoader(
     fun loadFrequencyDiagnosticData() {
         val audiometrics = audiometryRepository.findByAudiometryType(AudiometryType.SIGNALS.name)
         audiometrics.forEach { audiometry ->
+            val existTasks = audiometryTaskRepository.findByAudiometry(audiometry)
```
Dmitriy1Fokin, line 95: you can make a query in the database for the count of tasks

src/main/kotlin/com/epam/brn/service/load/AudiometryLoader.kt
```diff
@@ -92,6 +92,11 @@ class AudiometryLoader(
...
+            if (existTasks.size == 2)
+                return@forEach
+            if (existTasks.size > 2)
```
Dmitriy1Fokin, line 98: case when 1 existTasks is not necessary to check?

src/main/kotlin/com/epam/brn/service/load/AudiometryLoader.kt
ElenaSpb: done

src/main/kotlin/com/epam/brn/service/load/AudiometryLoader.kt
ElenaSpb: ok, added, it is inconsistent db status too

Dmitriy1Fokin (approved):

PR #2079, opened 2021-08-16
EPMLABSBRN-900 - BE: fix bug post headphones for current user: 403

goopnigoop:
[EPMLABSBRN-900](https://jira.epam.com/jira/browse/EPMLABSBRN-900)

- fix for the current api
- added new constraint for headphones table


ElenaSpb (approved):

ElenaSpb:
may be it is reasonable to add integration test on this bug?

goopnigoop:
> @goopnigoop , may be it is reasonable to add integration test on this bug?

ok, I'll add

PR #2082, opened 2021-09-02
2080 update mobile layout

Mikleprog:
EPMLABSBRN-2080

fix footer and header, exercise need to be done

### DIFF
--- a/frontend/app/components/exercise-type/styles.css
+++ b/frontend/app/components/exercise-type/styles.css
@@ -5,7 +5,8 @@
   border-radius: 10px;
 }
 
-.card {
+.card,
+.card-sm{
   align-items: center;
   border-radius: 10px;
   display: flex;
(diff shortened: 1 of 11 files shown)


frontend/app/components/exercise-type/styles.css
```diff
@@ -16,12 +17,21 @@
   width: 410px;
 }
 
+.card-sm{
```
lifeart, line 20: for local classes (not tailwind related) we have to use media-query expressions in css, instead of having multiple classes definition here. http://htmlbook.ru/css/value/media

frontend/app/components/footer/styles.css
```diff
@@ -9,7 +9,7 @@ footer {
 }
 
 a[rel='noopener'] {
-  display: block;
+  display: flex;
```
lifeart, line 12: should it be inline-block?

frontend/app/styles/tailwind.js
```diff
@@ -8,6 +8,9 @@ const {
 module.exports = {
   theme: {
     extend: {
+      fontSize: {
+        xss: '.5rem',
```
lifeart, line 12: we have to extends from builtin font-sizes, same as for colors: 
image

PR #2100, opened 2021-09-24
EPMLABSBRN-914 BE: Implement endpoints for adding doctors to patients

Shterneregen:


api-contract/api.raml
```diff
@@ -580,6 +580,19 @@ mediaType:
           body:
             example: "https://cloudprovider.bucket.com"
 
+# DOCTOR  ----------------------
+/doctor:
+    /patients:
+        put:
+            description: Add current doctor to patients
```
lifeart, line 587: patient -> belongs to doctor
doctor -> has many patient

patient could not control relationships
add patient to doctor will be 

'POST users/doctor_user_id/patients' { patient Id }'

remove patient from doctor will be

'DELETE users/doctor_id/patients/patient_id'

we don't need batch operations here or query params

Shterneregen: Ok! But why do we need `type: 'user'` in `add Patient to Doctor` request?

Shterneregen: What should return `get Patient Doctor` request? Doctor account? As far as I know, doctor is just a user with additional authority. We can get doctor info as for normal user. Or you suggest returning doctor DTO with a full list of patients?

src/main/kotlin/com/epam/brn/service/DoctorService.kt
```diff
@@ -0,0 +1,102 @@
...
+        val currentUser = userAccountService.getCurrentUser().toDto()
+        val patient = userAccountService.findUserById(patientId)
+        val doctor = userAccountService.findUserById(doctorId)
+
+        checkUserIsNotAdmin(patient, USING_ADMIN_ID_FOR_PATIENT_WARN)
+        checkUserIsNotAdmin(doctor, USING_ADMIN_ID_FOR_DOCTOR_WARN)
+
+        when {
+            !isAdmin(currentUser) && currentUser.id != doctorId -> {
+                throw IllegalArgumentException("It is forbidden to assign another doctor to patient")
```
ElenaSpb, line 21: It is forbidden to change data for another doctor.

src/main/kotlin/com/epam/brn/service/DoctorService.kt
```diff
@@ -0,0 +1,102 @@
...
+            }
+            !isDoctor(doctor) -> {
+                throw IllegalArgumentException("You are trying to assign patient to user that is not a doctor")
+            }
+            isDoctor(patient) -> {
+                throw IllegalArgumentException("[${patient.name}] is doctor, you cannot add doctor to another doctor")
+            }
+            patient.doctor != null -> throw IllegalArgumentException("The patient already have a doctor")
```
ElenaSpb, line 29: you delete previous one and then add other.

src/main/kotlin/com/epam/brn/service/DoctorService.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    fun removeDoctorFromPatientAsDoctor(doctorId: Long, patientId: Long) {
+        val currentUser = userAccountService.getCurrentUser().toDto()
+        val patient = userAccountService.findUserById(patientId)
+
+        checkUserIsNotAdmin(patient, USING_ADMIN_ID_FOR_PATIENT_WARN)
+
+        when {
+            !isAdmin(currentUser) && currentUser.id != doctorId -> {
+                throw IllegalArgumentException("It is forbidden to remove a patient from another doctor")
```
ElenaSpb, line 42: It is forbidden to change data for another doctor.

src/main/kotlin/com/epam/brn/service/DoctorService.kt
```diff
@@ -0,0 +1,102 @@
...
+            }
+            !isAdmin(currentUser) && patient.doctor != doctorId -> {
+                throw IllegalArgumentException("It is forbidden to remove another doctor from patient")
```
ElenaSpb, line 45: It is forbidden to change data for another doctor.

src/main/kotlin/com/epam/brn/service/DoctorService.kt
Shterneregen: One doctor can delete a patient from another doctor?

Issue #2098, opened 2021-09-21
[BE]:  Implement DELETE HeadPhones for current user functionality


rkonoplev:
As user I should have possibility to delete previous added 
on tech side: add field active , set it by default true,

on delete process - set field into false, not really delete, as we have history with it.

by default return only active headphones for user

PR #2101, opened 2021-09-24
[BE]: Implement DELETE HeadPhones for current user functionality

rkonoplev:
[EPMLABSBRN-913](https://github.com/Brain-up/brn/issues/2098)

**Description**:

As user I should have possibility to delete previous added
on tech side: add field active , set it by default true,

on delete process - set field into false, not really delete, as we have history with it.

by default return only active headphones for user

### DIFF
--- a/src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
+++ b/src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
@@ -11,6 +11,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.PatchMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
@@ -72,6 +73,12 @@ class UserDetailController(@Autowired val userAccountService: UserAccountService
     fun addHeadphonesToCurrentUser(@Validated @RequestBody headphones: HeadphonesDto) =
         ResponseEntity.status(HttpStatus.CREATED)
             .body(BaseSingleObjectResponseDto(data = userAccountService.addHeadphonesToCurrentUser(headphones)))
+
+    @DeleteMapping(value = ["/current/headphones"])
+    @ApiOperation("Delete headphones for the current user")
+    fun deleteHeadphonesForCurrentUser(@Validated @RequestBody headphones: HeadphonesDto) =
+        ResponseEntity.status(HttpStatus.OK)
+            .body(BaseSingleObjectResponseDto(data = userAccountService.deleteHeadphonesForCurrentUser(headphones)))
 
     @GetMapping(value = ["/{userId}/headphones"])
     @ApiOperation("Get all user's headphones")
--- a/src/main/kotlin/com/epam/brn/model/Headphones.kt
+++ b/src/main/kotlin/com/epam/brn/model/Headphones.kt
@@ -23,6 +23,7 @@ data class Headphones(
     var id: Long? = null,
     @NotNull
     var name: String = "",
+    var active: Boolean = true,
     @Enumerated(EnumType.STRING)
     var type: HeadphonesType,
     var description: String = "",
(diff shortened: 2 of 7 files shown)


Dmitriy1Fokin (changes requested):

out/production/resources/db/migration/V220211001_913.sql
```diff
@@ -0,0 +1 @@
+alter table headphones add column if not exists active boolean
```
Dmitriy1Fokin, line 1: package `out` is auto-generated. Move script to `src/main/resources/db/migration`

src/main/kotlin/com/epam/brn/controller/UserDetailController.kt
```diff
@@ -73,6 +74,16 @@ class UserDetailController(@Autowired val userAccountService: UserAccountService
         ResponseEntity.status(HttpStatus.CREATED)
             .body(BaseSingleObjectResponseDto(data = userAccountService.addHeadphonesToCurrentUser(headphones)))
 
+    @DeleteMapping(value = ["/current/headphones/{headphonesId}"])
+    @ApiOperation("Delete headphones by Id")
+    fun deleteHeadphonesForCurrentUser(
+        @PathVariable(value = "headphonesId") headphonesId:
+            Long
```
Dmitriy1Fokin, line 81: variable name and variable type must be on the same line

src/main/kotlin/com/epam/brn/model/Headphones.kt
```diff
@@ -23,6 +23,7 @@ data class Headphones(
     var id: Long? = null,
     @NotNull
     var name: String = "",
+    var active: Boolean?,
```
Dmitriy1Fokin, line 26: entity's field `active` must be non-nullable

src/main/kotlin/com/epam/brn/repo/UserAccountRepository.kt
```diff
@@ -26,4 +26,6 @@ interface UserAccountRepository : JpaRepository<UserAccount, Long> {
 
     @Query("select DISTINCT u FROM UserAccount u left JOIN FETCH u.authoritySet authorities left JOIN FETCH u.headphones where authorities.authorityName = :authorityName")
     fun findUsersAccountsByRole(authorityName: String): List<UserAccount>
+
+    fun deleteHeadphonesForCurrentUser(id: Long)
```
Dmitriy1Fokin, line 30: 1. why is this method needed
2. the method is incorrectly compiled. [Docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.query-creation)

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -98,7 +101,8 @@ class UserAccountServiceImpl(
 
     override fun getAllHeadphonesForUser(userId: Long) = headphonesService.getAllHeadphonesForUser(userId)
 
-    override fun getAllHeadphonesForCurrentUser() = getCurrentUser().headphones.map(Headphones::toDto).toSet()
+    override fun getAllHeadphonesForCurrentUser() = getCurrentUser()
+        .headphones.map(Headphones::toDto).filter { it.active!! }.toSet()
```
Dmitriy1Fokin, line 105: Better to filter first then convert

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -98,7 +101,8 @@ class UserAccountServiceImpl(
 
     override fun getAllHeadphonesForUser(userId: Long) = headphonesService.getAllHeadphonesForUser(userId)
```
Dmitriy1Fokin, line 102: and here it is not necessary to filter?

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -138,6 +142,13 @@ class UserAccountServiceImpl(
         return headphonesService.save(entityHeadphones)
     }
 
+    override fun deleteHeadphonesForCurrentUser(headphonesId: Long): Headphones {
```
Dmitriy1Fokin, line 145: why return the deleted headphones. Maybe it's better not to return anything?

Dmitriy1Fokin (changes requested):

PR #2103, opened 2021-10-04
2094 subgroup cards layout

Mikleprog:
2094 Subgroup cards layout

### DIFF
--- a/frontend/app/components/series-navigation/styles.css
+++ b/frontend/app/components/series-navigation/styles.css
@@ -33,6 +33,19 @@
   width: 168px;
   height: 168px;
 }
+@media (max-width: 500px){
+  .exercise-block__level {
+    width: 85px;
+    height: 85px;
+  }
+}
+
+@media (min-width: 501px) and (max-width: 1240px){
+  .exercise-block__level {
+    width: 115px;
+    height: 115px;
+  }
+}
 
 .exercise-block__status-icon {
   width: 20px;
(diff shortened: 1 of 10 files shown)


frontend/app/components/series-navigation/styles.css
```diff
@@ -33,6 +33,19 @@
   width: 168px;
   height: 168px;
 }
+@media (max-width: 500px){
```
lifeart, line 36: as I remember, we don't have breakpoints, matched for 500px
image

frontend/app/components/series-navigation/styles.css
```diff
@@ -33,6 +33,19 @@
...
+  .exercise-block__level {
+    width: 85px;
+    height: 85px;
+  }
+}
+
+@media (min-width: 501px) and (max-width: 1240px){
```
lifeart, line 43: here too, 640, 1280

Mikleprog:
Before: 
image
image
After:
image
image

frontend/app/components/series-navigation/styles.css
Mikleprog: I suggest using other values ​​for breakpoints, because 640 is too much for small screens. If we use 640 instead 500 it well be to much space, here : 
640
image
500
image

Because we do not fixed by tailwind breakpoints  using mediaquery,  and smaller values than 640  is better fit to small screens it is better also use other smaller breakpoints than 640px

PR #2121, opened 2021-12-02
Feature/epmlabsbrn 768 UI admin support moving to fire base auth

marklekin:
[EPMLABSBRN-768](https://jira.epam.com/jira/browse/EPMLABSBRN-768)

**Description**:
- Firebase added to admin UI
- environment variables for Firebase
- auth service updated
- authToken in local storage on successful login, removed on logout
- logout button to admin UI menu
- login error case snack bar
- auth interceptor uses firebase user accessToken
- fix: gender lowercase


frontend-angular/src/app/app-routing.module.ts
```diff
@@ -13,8 +13,8 @@ const routes: Routes = [
   },
   {
     path: '',
-    canLoad: [AuthAccessGuard],
-    canActivate: [AuthAccessGuard],
+    // canLoad: [AuthAccessGuard],
+    // canActivate: [AuthAccessGuard],
```
ElenaSpb, line 17: why do not you want to delete them at all?

marklekin: Good point! Restored the AuthAccessGuard. It was commented out for testing.

lifeart (approved):

PR #2137, opened 2022-01-11
EPMLABSBRN-968_task_readable

Mikleprog:
https://jira.epam.com/jira/browse/EPMLABSBRN-968

Changed text size, picture size


Mikleprog:
Not fully responsive yet.
Before: 
image
image
After:
image
  
image

frontend/app/components/task-player/single-simple-words/option.hbs
```diff
@@ -25,16 +25,16 @@
       )
     }}
   >
-    {{#if @answerOption.pictureFileUrl}}
```
lifeart, line 28: we need this "if" because some tasks may not have picture at all

frontend/app/components/task-player/single-simple-words/option.hbs
```diff
@@ -25,16 +25,16 @@
...
-      {{@answerOption.word}}
+    <div>
       <img
-        class="border-blue-50 border-2 rounded-lg"
+        class="sm:h-40 h-16 m-auto"
```
lifeart, line 30: let's keep border radius (it could be smaller)

frontend/app/components/task-player/single-simple-words/option.hbs
Mikleprog: Done.   But if do not works (early too)

frontend/app/components/task-player/single-simple-words/option.hbs
Mikleprog: Just add rounded-lg. Border was not add because without border the hole slide looks better especially if there is a white background.

Mikleprog:
Before: 
image
image
image

After: 
image
image
image

For small screen also should be added scroll and should not be used too long words (like видеооператор  or дельтапланеризм)

frontend/app/components/ui/tab-button/index.ts
```diff
@@ -11,13 +11,13 @@ interface UiTabButtonComponentArguments {
...
     const items = [
-      'focus:outline-none rounded-lg w-full uppercase h-12 overflow-hidden',
+      'focus:outline-none rounded-lg w-full uppercase h-12 overflow-hidden sm:text-sm text-xs',
     ];
-    if (this.args.small) {
-      items.push('text-xs');
-    } else {
-      items.push('text-sm');
-    }
+    // if (this.args.small) {
```
lifeart, line 16: why this options is removed?

lifeart (approved):

PR #2143, opened 2022-01-17
EPMLABSBRN-960 BE: update Spring, Kotlin to latest

Shterneregen:
[EPMLABSBRN-960](https://jira.epam.com/jira/browse/EPMLABSBRN-960)

- Spring Boot: 2.4.3 --> 2.5.8
- Kotlin: 1.4.20 --> 1.6.10
- build.gradle --> build.gradle.kts
- settings.gradle --> settings.gradle.kts
- Changes in UserAccountServiceTest and LopotkoProcessorTest were made because these tests got failed in Spring Boot 2.6.2, and I left it for the future

---

At the beginning, I wanted to use the latest stable Spring Boot 2.6.2, but I got an error while starting the app: 
Failed to start bean 'documentationPluginsBootstrapper'  
I investigated it a bit, and found that it's an issue between Spring Boot and Swagger

[Spring Boot - Swagger Documentation doesn't work](https://stackoverflow.com/questions/69108273/spring-boot-swagger-documentation-doesnt-work)
> The problem seems to be that Spring Boot switched from Ant Path Matching to using a PathPatternParser, see
> https://github.com/spring-projects/spring-boot/issues/24645
> https://github.com/spring-projects/spring-boot/issues/24805

[Failed to start bean 'documentationPluginsBootstrapper' in spring data rest](https://stackoverflow.com/questions/40241843/failed-to-start-bean-documentationpluginsbootstrapper-in-spring-data-rest)
> I got same issue using springfox-swagger2 and springfox-swagger-ui version(3.0.0), spring-boot version(2.6.2)
>The way to resolve this issue is by adding pathmatcher in application. properties or application.yml
> for application.properties:
> spring.mvc.pathmatch.matching-strategy = ANT_PATH_MATCHER

[Springboot 2.6.0 / Spring fox 3 - Failed to start bean 'documentationPluginsBootstrapper'](https://stackoverflow.com/questions/70036953/springboot-2-6-0-spring-fox-3-failed-to-start-bean-documentationpluginsboot)
> This problem's caused by a bug in Springfox. It's making an assumption about how Spring MVC is set up that doesn't always hold true. Specifically, it's assuming that MVC's path matching will use the Ant-based path matcher and not the PathPattern-based matcher. PathPattern-based matching has been an option for some time now and is the default as of Spring Boot 2.6.

> As described in Spring Boot 2.6's release notes, you can restore the configuration that Springfox assumes will be used by setting spring.mvc.pathmatch.matching-strategy to ant-path-matcher in your application.properties file. **Note that this will only work if you are not using Spring Boot's Actuator**. The Actuator always uses PathPattern-based parsing, irrespective of the configured matching-strategy. A change to Springfox will be required if you want to use it with the Actuator in Spring Boot 2.6 and later.


build.gradle.kts
```diff
@@ -0,0 +1,187 @@
...
+    id("org.sonarqube") version "3.0"
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation("io.kotest:kotest-assertions:$kotestAssertionsVerstion")
+    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
```
ElenaSpb, line 30: should it be updated to 11
or deleted?

Shterneregen: There is no `kotlin-stdlib-jdk11`
Some info:
[Why is there no kotlin-stdlib-jdk11?](https://stackoverflow.com/questions/65731542/why-is-there-no-kotlin-stdlib-jdk11)  
> TL;DR kotlin-stdlib-jdk8 can be successfully used with JDK 11.  

[What are plans for Kotlin runtimes after kotlin-stdlib-jdk8?](https://discuss.kotlinlang.org/t/what-are-plans-for-kotlin-runtimes-after-kotlin-stdlib-jdk8/20786)  
> I believe they target the JVM-bytecode of version 8, but Kotlin can be run with any higher JVM, since JVM is pretty backwards compatible most of the time.  

---

And I have some doubts about deleting, since after deleting 'kotlin-stdlib-jdk8' and 'kotlin-reflect' many integration tests were failed.
Moreover, when I'm generating a project using [start.spring.io](https://start.spring.io/) I also get 'kotlin-stdlib-jdk8' and 'kotlin-reflect' dependencies in build.gradle file (generated project: gradle, kotlin, java11)

ElenaSpb: as I understand we do not it if we will use
plugins {
    kotlin("jvm")
    kotlin("plugin.spring")
   ...
}
?

gradle.properties
```diff
@@ -1,8 +1,8 @@
 kotlin.code.style=official
 
-springBootVersion=2.4.3
+springBootVersion=2.5.8
 springDependencyVersion=1.0.11.RELEASE
-kotlinVersion=1.4.20
+kotlinVersion=1.6.10
 flywayVersion=6.5.5
 junitVersion=5.3.1
 mockkVersion=1.12.0
```
ElenaSpb, line 8: can we update it to latest too?
mockkVersion = 3.0.1?

build.gradle.kts
```diff
@@ -0,0 +1,194 @@
...
+    implementation("org.postgresql:postgresql")
+    implementation("org.flywaydb:flyway-core:$flywayVersion")
+
+    /* Firebase admin */
+    implementation("com.google.firebase:firebase-admin:7.3.0")
+
+    implementation("com.auth0:java-jwt:3.10.3")
+    implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-csv")
+
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.2")
```
ElenaSpb, line 56: let's update it to latest too

gradle.properties
Shterneregen: Ok, I'll do it, but the latest version is 1.12.2
image

build.gradle.kts
Shterneregen: done

Dmitriy1Fokin (approved):

PR #2199, opened 2022-03-22
Epmlabsbrn-1064 API for check if user has statistics

andrsam:
[EPMLABSBRN-1064](https://jira.epam.com/jira/browse/EPMLABSBRN-1064)

**API for check if user has statistics**:


src/main/kotlin/com/epam/brn/controller/StudyHistoryControllerV2.kt
```diff
@@ -26,4 +28,10 @@ class StudyHistoryControllerV2(
         @RequestParam("to", required = true) to: LocalDateTime
     ) = ResponseEntity.ok()
         .body(BaseResponse(data = studyHistoryService.getHistoriesForCurrentUser(from, to)))
+
+    @GetMapping("/user/{userId}/has/statistics")
```
Dmitriy1Fokin, line 32: Please add a new endpoint to the api-contract

Shterneregen:
I see that your pipeline is also failed with Metaspace cause (as mine), I solved it by adding `org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m` in `gradle.properties` file. You can try the same if you want.
image

src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
```diff
@@ -126,4 +126,10 @@ interface StudyHistoryRepository : CrudRepository<StudyHistory, Long> {
             "AND s.userAccount.id = :userId"
     )
     fun getDayStatistic(userId: Long, year: Int, month: Int, day: Int): List<StudyHistory>
+
+    @Query(
+        "select case when count (s) > 0 then true else false end " +
+            "from StudyHistory s where s.userAccount.id = :userId"
+    )
+    fun isUserHasStatistics(userId: Long): Boolean
```
Dmitriy1Fokin, line 134: you can make it easier:
1. `@Query("select count (s) > 0 from StudyHistory s where s.userAccount.id = :userId")`
or
2. ` fun existsByBlaBla(userId: Long): Boolean`

src/main/kotlin/com/epam/brn/controller/StudyHistoryControllerV2.kt
ElenaSpb: why not simplier 
/has/statistics -> /hasStatistics ?

PR #2237, opened 2022-06-22
2235 analytic fields

mironovandrei:
PR for #2235 
Replace mocks for workDayByLastMonth, firstDone, and lastDone with real data

PR #2247, opened 2022-07-14
#2237 BE: fix variable names in SQL query for first and last study

mironovandrei:


src/main/kotlin/com/epam/brn/repo/StudyHistoryRepository.kt
```diff
@@ -80,7 +80,7 @@ interface StudyHistoryRepository : CrudRepository<StudyHistory, Long> {
     fun findLastByUserAccountIdAndExerciseId(userId: Long, exerciseId: Long): StudyHistory?
 
     @Query(
-        "SELECT MIN(s.startTime) as firstVisit, MAX(s.startTime) as lastVisit FROM StudyHistory s WHERE user_id=:userId"
+        "SELECT MIN(s.startTime) as firstStudy, MAX(s.startTime) as lastStudy FROM StudyHistory s WHERE user_id=:userId"
     )
     fun findFirstAndLastVisitTimeByUserAccount(userId: Long?): FirstLastStudyView
```
Dmitriy1Fokin, line 85: I think the method should also be renamed

src/main/kotlin/com/epam/brn/service/impl/UserAnalyticsServiceImpl.kt
```diff
@@ -52,6 +52,7 @@ class UserAnalyticsServiceImpl(
             user.id?.let {
                 val firstAndLastVisitTimeByUserAccount =
                     studyHistoryRepository.findFirstAndLastVisitTimeByUserAccount(it)
+                println(firstAndLastVisitTimeByUserAccount)
```
Dmitriy1Fokin, line 55: I think it needs to be removed

PR #2260, opened 2022-08-01
EPMLABSBRN-1123 UI: Add "Spent time" and "Done exercises" columns

Mikleprog:
Add "Spent time" and "Done exercises" columns, add *ngIf conditions

### DIFF
--- a/frontend-angular/src/assets/i18n/en.json
+++ b/frontend-angular/src/assets/i18n/en.json
@@ -77,7 +77,9 @@
               "CurrentWeek": "Current week",
               "Name": "Name",
               "Progress": "Progress",
-              "StudyDaysInLastMonth": "Days in last month"
+              "StudyDaysInLastMonth": "Days in last month",
+              "SpentTime": "Spent time",
+              "DoneExercises": "Done exercises"
             },
             "NameColumn": {
               "CommonInfo [years, gender]": "{{ years }} years old",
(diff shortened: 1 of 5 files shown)


koylubaevnt (changes requested):
In file user.model.ts (src/app/modules/admin/models) need add new properties too.
export interface User:
1. "Spent time"
2. "Done exercises"

frontend-angular/src/assets/i18n/en.json
```diff
@@ -77,7 +77,9 @@
               "CurrentWeek": "Current week",
               "Name": "Name",
               "Progress": "Progress",
-              "StudyDaysInLastMonth": "Days in last month"
+              "StudyDaysInLastMonth": "Days in last month",
+              "SpentTime": "Spent time",
+              "DoneExercises": "Done exercises"
```
marklekin, line 82: Recommended to use 'Done' instead 'Done exercises'.
1) It's long and space is limited (if other colums have to be added...)
2) If only 1 in the column, it looks ugly to use plural 'exerciseS'.
+3) 'Exercises Done' would sound better (I don't know the translated version)
@ElenaSpb, @alextanlan please take a look, thanks!

Mikleprog: "Done exercises"  and "Spent time" will be displayed on two lines so an additional word doesn't  consume much space  and header looks pretty good even on a small screen. If new columns be added probably better to add Scroll.
image

marklekin (approved):

ElenaSpb (approved):

PR #2266, opened 2022-08-03
2255 add fields "spentTime" and "doneExercises" to response for user with analytics

mironovandrei:
1. Added fields "spentTime" and "doneExercises" to response for user with analytics.
2. Updated unit-tests


koylubaevnt (approved):

src/main/kotlin/com/epam/brn/dto/response/UserWithAnalyticsResponse.kt
```diff
@@ -22,5 +21,5 @@ data class UserWithAnalyticsResponse(
     var studyDaysInLastMonth: Int = 0, // amount of days in last month when user made any exercises
     var diagnosticProgress: Map<AudiometryType, Boolean> = mapOf(AudiometryType.SIGNALS to true), // todo fill by user
     var doneExercises: Int = 0, // for all time
-    var spentTime: Duration = Duration.ZERO, // spent time by doing exercises for all time
+    var spentTime: Long = 0L, // spent time by doing exercises for all time in milliseconds
```
ElenaSpb, line 24: why you change it in ms?
in this case some manipulation will be needed on UI part
why Duration type is not sutable?

prizrakbeckman (approved):

src/main/kotlin/com/epam/brn/dto/response/UserWithAnalyticsResponse.kt
mironovandrei: Because when we convert this object to JSON then string-representation of duration is nanoseconds*2(for example 10 seconds becomes 20000000000). 
And I don't see any code which would convert it to something better understandable for human.
So I've decided to change it to milliseconds as it's common way for time representation.
Let's decide what to do with that and I'll change this code according to our decision.

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
```diff
@@ -32,6 +33,8 @@ data class StudyHistory(
     var exercise: Exercise,
     var startTime: LocalDateTime,
     var endTime: LocalDateTime? = null,
+    var spentTime: Long? = endTime?.toInstant(ZoneOffset.UTC)?.toEpochMilli()
+        ?.minus(startTime.toInstant(ZoneOffset.UTC).toEpochMilli()),
```
ElenaSpb, line 37: actually milliseconds is extra for us,
let it be in seconds
spentTime -> spentSeconds
?

src/main/kotlin/com/epam/brn/dto/response/UserWithAnalyticsResponse.kt
ElenaSpb: as Kotlin provide us special type Duration for this purpose I think we should use it,
on UI part they will convert it into readable view the number that they will get. 
no?

mironovandrei: ok, that's reasonable
I'll use Duration here
but we need a task for front-end developers for convertation of this number

mironovandrei: done

src/main/kotlin/com/epam/brn/model/StudyHistory.kt
mironovandrei: done

ElenaSpb (approved):

PR #2267, opened 2022-08-05
2265 remove deprecated methods

koylubaevnt:
1. Removed deprecated methods:
  * /admin/histories
  * /status-history/histories
  * /admin/study/week
  * /statistics/study/week
  * /admin/study/year
  * /statistics/study/year
2. Change frontend code to use v2 similar methods
3. Remove unused and commented code

### DIFF
--- a/frontend/app/adapters/user-weekly-statistics.ts
+++ b/frontend/app/adapters/user-weekly-statistics.ts
@@ -3,7 +3,7 @@ import ApplicationAdapter from './application';
 
 export default class UserWeeklyStatisticsAdapter extends ApplicationAdapter {
   pathForType() {
-    return 'statistics/study/week';
+    return 'v2/statistics/study/week';
   }
 
   sortQueryParams(query: { from: DateTime; to: DateTime }): {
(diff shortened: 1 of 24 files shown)


frontend/app/utils/custom-timeout.js
```diff
@@ -1,7 +1,7 @@
 import { timeout } from 'ember-concurrency';
 import Ember from 'ember';
 
-export default async function customTimeout(ms = 16) {
-  const delay = Ember.testing ? 1 : ms;
+export default async function customTimeout(ms = 16, useMsValueForTest = false) {
```
lifeart, line 4: we don't need this extra flag, we rely on `Ember.testing`, it's `true` only during tests

frontend/app/adapters/user-weekly-statistics.ts
```diff
@@ -3,23 +3,23 @@ import ApplicationAdapter from './application';
...
   }
 
   sortQueryParams(query: { from: DateTime; to: DateTime }): {
     from: string;
     to: string;
   } {
     const newQuery = {
-      from: query.from.toUTC().toFormat('yyyy-MM-dd'),
-      to: query.to.toUTC().toFormat('yyyy-MM-dd'),
+      from: query.from.toUTC().toFormat('yyyy-MM-dd\'T\'HH:mm:ss'),
```
lifeart, line 14: I think we could use `toISOString()` here

koylubaevnt: No, we can't. 
Because backend allow only format without TIMEZONE, but toISO() generates date in string format with timezone. And we got answer from back with error: Could not parse ...

frontend/app/utils/custom-timeout.js
koylubaevnt: I forgot to revert it. It's just tried to make longer timeout for test. Reverted it

PR #2293, opened 2022-08-24
2285: added POST & PUT endpoints for contributors

RudzMaksim:


src/main/kotlin/com/epam/brn/service/ContributorServiceImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+        contributor.name = dto.name
+        contributor.nameEn = dto.nameEn
+        contributor.description = dto.description
+        contributor.descriptionEn = dto.descriptionEn
+        contributor.company = dto.company
+        contributor.companyEn = dto.companyEn
+        contributor.contribution = dto.contribution!!
+        contributor.pictureUrl = dto.pictureUrl
+        contributor.type = dto.type!!
+        contributor.contacts.clear()
```
RudzMaksim, line 42: Per our discussion, I applied the approach to clear all contacts and add ones from user input.
It simplified 'update' operation.
Alternatively, we would need to separate 'Create/Update' dto objects for contacts and add extra logic for update trying to understand what was added/updated/removed

src/main/kotlin/com/epam/brn/config/WebSecurityBasicConfiguration.kt
```diff
@@ -46,6 +47,9 @@ class WebSecurityBasicConfiguration(
             .antMatchers("/cloud/upload").hasAuthority(ROLE_ADMIN.name)
             .antMatchers("/cloud/folders").hasAuthority(ROLE_ADMIN.name)
             .antMatchers("/doctors/**").hasAnyAuthority(ROLE_ADMIN.name, ROLE_DOCTOR.name)
+            .antMatchers(HttpMethod.GET, "/contributors/**").permitAll()
```
RudzMaksim, line 50: Per our discussion, I implemented the necessary security config via antMatchers.
Changes to revise our approach will be addressed via a separate issue.

src/main/kotlin/com/epam/brn/service/ContributorServiceImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.epam.brn.dto.response.ContributorResponse
+import com.epam.brn.enums.ContributorType
+import com.epam.brn.exception.EntityNotFoundException
+import com.epam.brn.repo.ContributorRepository
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import java.util.stream.Collectors
+
+@Service
+class ContributorServiceImpl(
```
RudzMaksim, line 13: I renamed this class as simple implementation of interface
The existing 'getContributors' method was not related to GitHub as new methods do.
So I decided that 'Impl' class is a good enough name here.

src/main/kotlin/com/epam/brn/dto/request/contributor/ContactRequest.kt
```diff
@@ -0,0 +1,21 @@
...
+import javax.validation.constraints.NotBlank
+import javax.validation.constraints.NotNull
+
+data class ContactRequest(
+    @field:NotNull
+    val type: ContactType?,
+
+    @field:NotBlank
+    @field:Length(max = 255)
+    val value: String?
```
RudzMaksim, line 15: It was agreed with @ElenaSpb  that we're fine to proceed without validations for contacts (email, phone number).
It would be easier to do with first-level fields via annotations but with such structure like this it will be more complicated so we agreed to skip it for now with the assumption that our admins are adequate people and won't submit obviously wrong values.

src/main/kotlin/com/epam/brn/dto/request/contributor/ContributorRequest.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.hibernate.validator.constraints.Length
+import javax.validation.Valid
+import javax.validation.constraints.NotBlank
+import javax.validation.constraints.NotNull
+import javax.validation.constraints.Positive
+
+data class ContributorRequest(
+    @field:NotBlank
+    @field:Length(max = 255)
+    val name: String?,
```
ElenaSpb, line 14: name, description,
nameEn, descriptionEn 
are not nullable type
String? -> String

https://github.com/Brain-up/brn/wiki/Kotlin-request-dto-validation-with-annotations as I remember with String validation works correct

ElenaSpb:
ok, thanks, Maksim!
waiting tests

src/main/kotlin/com/epam/brn/dto/request/contributor/ContributorRequest.kt
RudzMaksim: I saw it in the docs and I tried to use the following approach:
@field:NotBlank(message = "{validation.field.name.blank}")
@field:Length(max = 255)
 val name: String
However, I saw the ugly Kotlin error message instead of the validation message, so I made String fields nullable as well.

Regarding 'En' fields, I made them nullable because I thought that if we speak about specialists they might not have any English values, it will be info just in Russian. But if we want to force admin to provide both Russian and English values for specialists, I'll add not empty validations

RudzMaksim: I made 'En' fields required, we'll be able to adjust it as needed later
I left String fields nullable to avoid the Kotlin error message.

AnatolVoluznev (approved):

PR #2296, opened 2022-08-30
2284 BE: typed responses

AnatolVoluznev:


ElenaSpb:
could you please use new Response in ContributorController too?

src/main/kotlin/com/epam/brn/dto/response/BaseResponse.kt
```diff
@@ -1,11 +1,19 @@
...
+
+@Deprecated("use Response")
 data class BaseResponse(
     val data: List<Any> = emptyList(),
     val errors: List<Any> = emptyList(),
     val meta: List<Any> = emptyList()
 )
 
+@Deprecated("use Response")
 data class BaseSingleObjectResponse(
```
ElenaSpb, line 17: if this deprecated classes are not use anywhere so it is reasonable to delete them?

AnatolVoluznev: yes, removed

ElenaSpb (approved):

PR #2300, opened 2022-09-21
2295 BE: redesign API around resources, enhanced swagger docs with tags …

RudzMaksim:
2295 BE: redesign API around resources, enhanced swagger docs with tags


src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
```diff
@@ -0,0 +1,16 @@
+package com.epam.brn.enums
+
+enum class AuthorityType(val roleName: String) {
```
RudzMaksim, line 3: ENUM members cannot be used as annotation parameters for 'RolesAllowed'.
So I tried to restructure this class a little to come up with constants.
Actually, we deal in our app with two aspects: authorities & roles so I separated it by these changes
I left RoleConstants class inside this file as I haven't found any class for constants.
Please, let me know whether it's fine or I need to move it anywhere else.

src/main/kotlin/com/epam/brn/controller/StudyHistoryControllerV2.kt
```diff
@@ -13,21 +16,29 @@ import org.springframework.web.bind.annotation.RequestMapping
...
-    @ApiOperation("Get current user's study histories for period from <= startTime < to. Where from and to are dates in ISO format")
+    @ApiOperation("Get user's study histories for period from <= startTime <= to where startTime is a date in ISO date time format")
     fun getHistories(
+        @RequestParam("userId") userId: Long?,
         @RequestParam("from", required = true) from: LocalDateTime,
         @RequestParam("to", required = true) to: LocalDateTime
-    ) = ResponseEntity.ok()
-        .body(BaseResponse(data = studyHistoryService.getHistoriesForCurrentUser(from, to)))
+    ) {
+        val result = if (userId != null && AuthorityService.hasAuthority(AuthorityType.ROLE_ADMIN)) {
```
RudzMaksim, line 35: Example of the combination endpoints together which we discussed.
We have several endpoints where we do the same action either for any user or for a current one.
It's more compact to combine such endpoints together and have authority logic inside.

src/main/kotlin/com/epam/brn/config/OperationRolesAllowedReader.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.stereotype.Component
+import springfox.documentation.spi.DocumentationType
+import springfox.documentation.spi.service.OperationBuilderPlugin
+import springfox.documentation.spi.service.contexts.OperationContext
+import springfox.documentation.spring.web.DescriptionResolver
+import springfox.documentation.swagger.common.SwaggerPluginSupport
+import javax.annotation.security.RolesAllowed
+
+@Component
+class OperationRolesAllowedReader(
```
RudzMaksim, line 14: It turned out to be not very complicated to implement the reader to add 'Allowed Roles' info to our Swagger docs.
This is how it looks after the changes:
image
I've assigned the corresponding ticket on me and will close it together with this one

src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
```diff
@@ -0,0 +1,16 @@
...
+    ROLE_USER(RoleConstants.USER),
+    ROLE_DOCTOR(RoleConstants.DOCTOR);
+}
+
+class RoleConstants {
+    companion object {
+        const val ADMIN = "ADMIN"
+        const val USER = "USER"
+        const val DOCTOR = "DOCTOR"
+        const val ANONYMOUS = "ANONYMOUS"
```
RudzMaksim, line 14: This is the role that comes from Spring by default for not authorized users.
We need it for public endpoints (getContributors for now)

RudzMaksim:
This is all code changes we'll need for this redesign.
I created a Draft PR to have the ability to start reviewing these changes.
Remaining scope:
-Unit Tests
-Script to assign additional roles to the ADMIN user
-Change URLs for FE

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
```diff
@@ -84,7 +84,7 @@ describe('AdminApiService', () => {
 
   it('should call get exercises by subgroup id', () => {
     let exercise: Exercise[] | undefined;
-    const url = `${baseUrl}/exercises?subGroupId=${subGroupId}`;
+    const url = `${baseUrl}/exercises/search?subGroupId=${subGroupId}`;
```
ElenaSpb, line 87: why did you add 'search' in end-points?
semantically it is obvious by default get-search

src/main/kotlin/com/epam/brn/auth/AuthorityService.kt
```diff
@@ -9,4 +10,5 @@ interface AuthorityService {
     fun findAuthorityByAuthorityName(authorityName: String): Authority
     fun findAll(): List<AuthorityResponse>
     fun save(authority: Authority): Authority
+    fun hasAuthority(authorityType: AuthorityType): Boolean
```
ElenaSpb, line 13: who has authority?
not very clear what it is about

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
RudzMaksim: Because we already have the '/exercises' endpoint for the regular user so I had to come up with something unique

src/main/kotlin/com/epam/brn/auth/AuthorityService.kt
RudzMaksim: renamed to 'isCurrentUserHasAuthority'

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
ElenaSpb: so if end-points are separated base on role 
if would be more readable to see this semantic in url path instead of 'search' appendix. ?

src/main/kotlin/com/epam/brn/config/OperationRolesAllowedReader.kt
ElenaSpb: which is nice!

src/main/kotlin/com/epam/brn/config/OperationRolesAllowedReader.kt
```diff
@@ -0,0 +1,44 @@
...
+    @Autowired private val descriptions: DescriptionResolver
+) : OperationBuilderPlugin {
+
+    private val log = logger()
+
+    override fun apply(context: OperationContext) {
+        try {
+            var allowedRoles: Array<String>? = null
+            val rolesAllowedAnnotation = context.findAnnotation(RolesAllowed::class.java)
+            if (rolesAllowedAnnotation.isPresent) {
```
ElenaSpb, line 24: we do not use {}
in case with one code line in condition
https://github.com/Brain-up/brn/wiki/Code-Style

src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
```diff
@@ -0,0 +1,15 @@
+package com.epam.brn.enums
+
+enum class AuthorityType(val roleName: String) {
+    ROLE_ADMIN(RoleConstants.ADMIN),
+    ROLE_USER(RoleConstants.USER),
+    ROLE_DOCTOR(RoleConstants.DOCTOR);
+}
+
+class RoleConstants {
```
ElenaSpb, line 9: maybe name it 
RoleConstants -> BrnRoles?

src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
```diff
@@ -0,0 +1,15 @@
+package com.epam.brn.enums
+
+enum class AuthorityType(val roleName: String) {
```
ElenaSpb, line 3: let's name it 
AuthorityType -> BrnAuthorities 
to not confuse it with some framework/java classes
?

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
RudzMaksim: In this case, it's not so easy to make one endpoint because they have identical query parameters, there is no 'userId' parameter like for other cases.
As for the 'search' appendix, we can name this endpoint like '/exercises/admin' but I would propose to have the verb appendix as it looks to me more natural.
Anyway, I'll be fine with any option. If you want to change this endpoint's name, please give me the option of how you want this endpoint to be named and I'll update it.

src/main/kotlin/com/epam/brn/config/OperationRolesAllowedReader.kt
RudzMaksim: Got it. Fixed.

src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
RudzMaksim: Well, AuthorityType is a unique name, we don't have any conflicts with frameworks or Java.
I would keep it simple as we don't use the application prefix for the 'enums' package.
For example, we have 'Locale' enum. This name can be actually confusing because it's used by Java and frameworks, but I would say that we're fine as packages help to differentiate such things.
But, of course, I can name it as BrnAuthorities if you prefer this option.
Just wanted to emphasize the point that if we introduce application prefix for classes we introduce inconsistency and complexity for developers to answer the question of when it should be used and when it shouldn't

src/main/kotlin/com/epam/brn/enums/AuthorityType.kt
RudzMaksim: See my comment regarding 'AuthorityType'
I'm fine to name it anyhow but want to confirm that we really need it.

src/main/kotlin/com/epam/brn/controller/ExerciseController.kt
```diff
@@ -17,11 +23,40 @@ import org.springframework.web.bind.annotation.RequestBody
...
+    @ApiOperation("Create new exercise for existing subgroup")
+    @RolesAllowed(RoleConstants.ADMIN)
+    fun createExercise(
+        @ApiParam(value = "Exercise data", required = true)
+        @Valid @RequestBody exerciseCreateDto: ExerciseCreateDto
+    ): ResponseEntity<Response<ExerciseDto>> =
+        ResponseEntity.status(HttpStatus.CREATED)
+            .body(Response(data = exerciseService.createExercise(exerciseCreateDto)))
+
+    @GetMapping("/search")
```
ElenaSpb, line 49: let it be just 
@GetMapping

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
ElenaSpb: or maybe name end-point with all exercises for admin like
"/exercises/general"
"/exercises/all"
?

RudzMaksim: If 'search' postfix is not an option.
I would propose to move admin endpoint to SubGroupController then with pathVariable.
I made a commit for this solution

frontend-angular/src/app/modules/admin/services/api/admin-api.service.spec.ts
```diff
@@ -176,7 +176,7 @@ describe('AdminApiService', () => {
 
   it('should call get users', () => {
     let usersData: User[] | undefined;
-    const url = `${baseUrl}/users`;
+    const url = `${baseUrl}/users/search`;
```
ElenaSpb, line 179: why 'search' appendix appear here too?
let's avoid such extra not semantic clear code

RudzMaksim: We have the path conflict for Users API as well.
We have a general search endpoint and search by name where 'name' is RequestParam.
I dived into the endpoint related to search by name and couldn't find any usage of this endpoint.
So I dropped it from our API but please let me know if this endpoint is needed, then we'll need to think about other options.

ElenaSpb:
also we have this api description
https://github.com/Brain-up/brn/blob/master/api-contract/api.raml 

could you please fix it according current new state

ElenaSpb (approved):

PR #2302, opened 2022-09-22
EPMLABSBRN-2299 [BE & FrontEnd] Refactoring: merge controllers StudyH…

Isink21:
…istoryController and StudyHistoryControllerV2 into one class

**Description**:


ElenaSpb:
This task is not only about backend - it is general task about front and back.
now this changes will broke the platform

so you need to find all usages of this end-points and change them too

Issue #2286, opened 2022-08-19
[BE] Prepare sql script to addd surdo/speech specialists into Contributor table 


ElenaSpb:
[BE] Prepare sql script to fill surdo/speech specialists into application

input: https://github.com/Brain-up/brn/blob/master/frontend/app/services/persons.ts 
Contributor data model https://github.com/Brain-up/brn/wiki/Data-Models

PR #2331, opened 2022-10-12
2286 [BE] Prepare sql script to addd surdo/speech specialists into Contributor table #2286

koylubaevnt:
* Made migration script for load SPECIALIST
* Fix some problem in model Contributor: contribution field can be null


src/main/kotlin/com/epam/brn/model/Contributor.kt
```diff
@@ -32,7 +32,7 @@ data class Contributor(
     @Enumerated(EnumType.STRING)
     var type: ContributorType = ContributorType.DEVELOPER,
     var pictureUrl: String? = null,
-    var contribution: Long = 0,
+    var contribution: Long? = null,
```
ElenaSpb, line 35: why did you brake Kotlinish stile?
the intention is to have not nullable types.

here it is logical to have Long type with 0 as default.

koylubaevnt: Because in Database this value can be null. 
Will we replace to non nullable values?

ElenaSpb: semantically it is non null value
let's replace in DB too. ?

koylubaevnt: OK

koylubaevnt: I fixed it. Please review again

ElenaSpb (approved):

PR #2335, opened 2022-10-13
2324: update raml

RudzMaksim:
update raml

### DIFF
--- a/api-contract/api.raml
+++ b/api-contract/api.raml
@@ -204,9 +204,31 @@ annotationTypes:
       200:
         body:
           example: !include samples/subgroup_by_series.json
+  post:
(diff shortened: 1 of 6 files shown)


api-contract/api.raml
```diff
@@ -768,102 +711,4 @@ annotationTypes:
         responses:
           200:
             body:
-              example: !include samples/study_histories_dto_response.json
-    /study/day:
```
ElenaSpb, line 772: I did not find this end-point in new version.
?

RudzMaksim: Yeah. Thank you, added.

ElenaSpb (approved):

PR #2344, opened 2022-10-16
2311 be investigation refactoring jpa data classes WIP

AnatolVoluznev:

**Description**:

### DIFF
--- a/src/main/kotlin/com/epam/brn/model/Authority.kt
+++ b/src/main/kotlin/com/epam/brn/model/Authority.kt
@@ -8,7 +8,7 @@ import javax.persistence.GenerationType
 import javax.persistence.Id
 
 @Entity
-data class Authority(
+class Authority(
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     var id: Long? = null,
--- a/src/main/kotlin/com/epam/brn/model/Headphones.kt
+++ b/src/main/kotlin/com/epam/brn/model/Headphones.kt
@@ -17,7 +17,7 @@ import javax.validation.constraints.NotNull
 
 @Entity
 @Table(uniqueConstraints = [UniqueConstraint(columnNames = ["name", "userId"])])
-data class Headphones(
+class Headphones(
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     var id: Long? = null,
--- a/src/main/kotlin/com/epam/brn/model/Headphones.kt
+++ b/src/main/kotlin/com/epam/brn/model/Headphones.kt
@@ -54,13 +54,6 @@ class Headphones(
         return true
     }
 
-    override fun hashCode(): Int {
-        var result = id?.hashCode() ?: 0
-        result = 31 * result + name.hashCode()
-        result = 31 * result + type.hashCode()
-        return result
-    }
-
     override fun toString(): String {
         return "Headphones(id=$id, name='$name', type=$type, description='$description')"
     }
--- a/src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
+++ b/src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
@@ -552,3 +552,15 @@ internal class UserAccountServiceTest {
         }
     }
 }
+
+private fun UserAccount.copy() = UserAccount(
+    id = this.id,
+    fullName = this.fullName,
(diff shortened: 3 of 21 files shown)


src/main/kotlin/com/epam/brn/model/Authority.kt
```diff
@@ -23,4 +23,21 @@ data class Authority(
...
+        if (javaClass != other?.javaClass) return false
+
+        other as Authority
+
+        if (id != other.id) return false
+        if (authorityName != other.authorityName) return false
+        return true
+    }
+
+    override fun hashCode(): Int {
```
koylubaevnt, line 38: I think this should be delete

src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
```diff
@@ -539,3 +539,15 @@ internal class UserAccountServiceTest {
         }
     }
 }
+
+private fun UserAccount.copy() = UserAccount(
```
koylubaevnt, line 543: Maybe better to move this func to Util class, which create in util package?

src/test/kotlin/com/epam/brn/upload/csv/audiometrySpeech/LopotkoRecordProcessorTest.kt
```diff
@@ -194,3 +194,13 @@ internal class LopotkoRecordProcessorTest {
         )
     }
 }
+
+private fun AudiometryTask.copy(id: Long? = this.id, answerOptions: MutableSet<Resource> = this.answerOptions) = AudiometryTask(
```
koylubaevnt, line 198: Maybe better to move this func to Util class, which create in util package?

koylubaevnt (changes requested):

src/test/kotlin/com/epam/brn/upload/csv/audiometrySpeech/LopotkoRecordProcessorTest.kt
ElenaSpb: maybe better to move this function into AudiometryTask class
as it may be useful for others?

src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
ElenaSpb: maybe better to move this function into UserAccount class
as it may be useful for others?

src/test/kotlin/com/epam/brn/upload/csv/audiometrySpeech/LopotkoRecordProcessorTest.kt
AnatolVoluznev: I don't think `.copy` method in entities will be useful anywhere other than tests, I like solution with util package in test module

src/main/kotlin/com/epam/brn/model/Authority.kt
AnatolVoluznev: removed

src/test/kotlin/com/epam/brn/service/UserAccountServiceTest.kt
AnatolVoluznev: fixed

src/main/kotlin/com/epam/brn/model/Headphones.kt
```diff
@@ -41,26 +41,6 @@ data class Headphones(
         userAccount = userAccount?.id
     )
 
-    override fun equals(other: Any?): Boolean {
```
ElenaSpb, line 44: why in this class equals is also deleted?

src/main/kotlin/com/epam/brn/dto/response/TaskResponse.kt
```diff
@@ -11,6 +11,6 @@ data class TaskResponse(
     val exerciseMechanism: ExerciseMechanism = exerciseType.toMechanism(),
     val name: String? = "",
     val serialNumber: Int? = 0,
-    val answerOptions: Set<ResourceResponse> = HashSet(),
+    val answerOptions: List<ResourceResponse> = mutableListOf(),
```
ElenaSpb, line 14: as we agree to change all sets into list
please check all other places and change them please
image

Issue #2336, opened 2022-10-14
[BE Admin] Design and implement api for searching exercises for word


Mikleprog:
![Uploading image.png…]()

PR #2355, opened 2022-10-23
[WIP] 2336 Design and implement api for searching exercises for word

mironovandrei:
#2336 
[BE Admin] Design and implement api for searching exercises for word

Implemented endpoint "/api/exercises" which returns list of exercises available for current user or all the exercises if current user is admin.

### DIFF
--- a/src/main/kotlin/com/epam/brn/dto/request/contributor/ContributorRequest.kt
+++ b/src/main/kotlin/com/epam/brn/dto/request/contributor/ContributorRequest.kt
@@ -11,7 +11,7 @@ import javax.validation.constraints.Positive
 data class ContributorRequest(
     @field:NotBlank
     @field:Length(max = 255)
-    val name: String?,
+    val name: String,
 
     @field:NotBlank
     @field:Length(max = 255)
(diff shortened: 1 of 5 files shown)


api-contract/api.raml
```diff
@@ -258,11 +258,11 @@ annotationTypes:
 /exercises:
   description: Contains actions over exercises
   get:
-    description: Get exercises for subgroup with tasks. If called by current user, availability calculation is included
+    description: Get exercises for subgroup with tasks or all the exercises if subgroup is not provided. If called by current user, availability calculation is included
```
ElenaSpb, line 261: now we have thousands of exercises and such end-point with all exercises is not what we need here. it is not reasonable

src/main/kotlin/com/epam/brn/dto/ExerciseDto.kt
```diff
@@ -6,6 +6,7 @@ import java.time.ZoneOffset
 
 data class ExerciseDto(
     var seriesId: Long?,
+    var subGroupName: String? = null,
```
ElenaSpb, line 9: it  also should be new ExerciseWithWordsResponse
with words in response

PR #2357, opened 2022-11-01
#2336 BE: implemented endpoint for getting of exercises containing sp…

mironovandrei:
https://github.com/Brain-up/brn/issues/2336
[BE Admin] Design and implement api for searching exercises by word

Implemented endpoint "/api/exercises/byWord?word=..." which returns list of exercises containing specified word

### DIFF
--- a/src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
+++ b/src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
@@ -11,7 +11,7 @@ data class ExerciseWithWordsResponse(
     val changedWhen: LocalDateTime = LocalDateTime.now(ZoneOffset.UTC),
     val playWordsCount: Int? = 1,
     val wordsColumns: Int? = 3,
-    val words: List<String>,
+    val words: Map<Long, String> = emptyMap(),
     val subGroupName: String?,
     val seriesName: String?
 )
--- a/src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
+++ b/src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
@@ -1,0 +1,17 @@
+package com.epam.brn.dto.response
+
+import java.time.LocalDateTime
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/epam/brn/service/ExerciseService.kt
```diff
@@ -154,6 +155,11 @@ class ExerciseService(
         return subGroupExercises.map { it.toDto() }
     }
 
+    fun findExerciseByWord(word: String): List<ExerciseWithWordsResponse> {
```
ElenaSpb, line 158: findExerciseByWord ->
findExercisesByWord

src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+
+data class ExerciseWithWordsResponse(
+    val id: Long?,
+    val name: String,
+    val active: Boolean = true,
+    val changedBy: String? = "InitialDataLoader",
+    val changedWhen: LocalDateTime = LocalDateTime.now(ZoneOffset.UTC),
+    val playWordsCount: Int? = 1,
+    val wordsColumns: Int? = 3,
+    val words: List<String>,
```
ElenaSpb, line 14: also it should return a Resource.id
as our use case is to 
1.find exercises with wrong words
2.fix them (here we need resource.id to make call to back for edit)
so maybe it should be a map<resourceId, word>
?

src/main/kotlin/com/epam/brn/service/ExerciseService.kt
mironovandrei: renamed

src/main/kotlin/com/epam/brn/dto/response/ExerciseWithWordsResponse.kt
mironovandrei: ok
added Resource.id as a map <resourseId, word>

ElenaSpb (approved):

PR #2365, opened 2022-11-03
2309: Authority to Role migration

RudzMaksim:

**Description**:
-migrated Authority to Role dropping 'ROLE_' prefix
-'ROLE_' prefix remains only in one place (CustomUserDetails) due to specifics of how Spring Security authorities should be named
-the corresponding changes were done to BE, FE and database


RudzMaksim:
@ElenaSpb , it's BE + DB changes.
Most likely I'll need to update FE as well, will check it in the upcoming days, but BE & DB parts are ready fo review.

ElenaSpb:
> 

please add readable description and correct MR title 
according our code standarts
https://github.com/Brain-up/brn/wiki/Coding-Standards 5 point

src/main/kotlin/com/epam/brn/service/RoleService.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.epam.brn.dto.response.UserAccountResponse
+import com.epam.brn.model.Role
+
+interface RoleService {
+    fun findById(id: Long): Role
+    fun findByName(name: String): Role
+    fun findAll(): List<Role>
+    fun save(role: Role): Role
+    fun isUserHasRole(role: String): Boolean
+    fun isUserHasRole(user: UserAccountResponse, role: String): Boolean
```
ElenaSpb, line 12: please, add here 
isCurrentUserAdmin() as we can use it in many places
let be this logic here
?

src/main/kotlin/com/epam/brn/service/DoctorService.kt
```diff
@@ -1,12 +1,15 @@
...
 
 import com.epam.brn.dto.response.UserAccountResponse
-import com.epam.brn.enums.AuthorityType
+import com.epam.brn.enums.BrnRole
 import org.apache.logging.log4j.kotlin.logger
 import org.springframework.stereotype.Service
 
 @Service
-class DoctorService(private val userAccountService: UserAccountService) {
+class DoctorService(
```
ElenaSpb, line 9: logically it should be SpecialistService now according our naming

src/main/resources/application-dev.properties
```diff
@@ -1,5 +1,5 @@
-spring.datasource.username=${POSTGRES_USER:admin}
-spring.datasource.password=${POSTGRES_PASSWORD:admin}
+spring.datasource.username=${POSTGRES_USER:postgres}
+spring.datasource.password=${POSTGRES_PASSWORD:postgres}
```
ElenaSpb, line 2: by mistake?

src/main/kotlin/com/epam/brn/service/RoleService.kt
RudzMaksim: Well, I would leave it as I did:
-two overloaded methods
-one to check a role for a current user
-another to check a role for a specific user
I see it as more consistent and ended logic, having both a unified method and one specific for admin looks overloaded to me.
And I don't see a big difference between **isCurrentUserAdmin()** & **isUserHasRole(BrnRole.ADMIN)** calls.
But I can return this method if you think that it still will be better

src/main/kotlin/com/epam/brn/service/DoctorService.kt
RudzMaksim: Yeah, it's not related to my changes but I can make this renaming in the scope of my MR.

src/main/resources/application-dev.properties
RudzMaksim: yes, sorry
will fix

src/main/kotlin/com/epam/brn/service/DoctorService.kt
RudzMaksim: Actually, there are a lot of places where 'doctor' wording is still used.
I would propose it to move to another small ticket as this is not part of my scope

src/main/kotlin/com/epam/brn/service/RoleService.kt
ElenaSpb: isCurrentUserAdmin() is what we really need I think
I should not think what roles he should have, we just want to check if he isAdmin that's it
please leave isCurrentUserAdmin()

ElenaSpb (approved):

PR #2443, opened 2023-01-18
2407 [BE] Load contributors from github project repositories

andrsam:
##2407

**Load contributors from github project repositories**:


src/main/kotlin/com/epam/brn/service/ContributorServiceImpl.kt
```diff
@@ -55,17 +55,18 @@ class ContributorServiceImpl(
     }
 
     @Transactional
-    override fun createOrUpdateByGitHubUser(gitHubUser: GitHubUser): Contributor {
+    override fun createOrUpdateByGitHubUser(gitHubUser: GitHubUser, repositoryName: String): Contributor {
         val existContributor = contributorRepository.findByGitHubUser(gitHubUser)
         return existContributor
-            ?.updateByGitHubUser(gitHubUser)
-            ?: createContributor(gitHubUser)
+            ?.updateByGitHubUser(gitHubUser, repositoryName)
```
ElenaSpb, line 61: as repositoryName is semantically a gitHubUser property 
why do not you want to add in on it level?

andrsam: it's because of single github user can belong multiple repositories

ElenaSpb: so it can be a list of repos?

andrsam: there is a several solutions for this problem:

1. create two tables - one for repos and link table for contributors
2. not create a separate entity for storing repo name, create the table for storing contributor id and repository name
3. store the repository names as comma separated values, convert it to set, and if the new reposotory name doesn't contain in the list, add it to set and convert back to string

ElenaSpb: why do not you want to have one repo field in contributor
it is how we organize now

andrsam: in this PR there is a field repositoryName in contributor

ElenaSpb (approved):

ElenaSpb:
Great job, thank you!

PR #2444, opened 2023-01-18
2406: add last visit field and set it when user logs in

chackomathew:
##2406


src/main/kotlin/com/epam/brn/model/UserAccount.kt
```diff
@@ -53,6 +53,7 @@ data class UserAccount(
     @OneToMany(mappedBy = "userAccount", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
     var headphones: MutableSet<Headphones> = hashSetOf()
 ) {
+    var lastVisit: LocalDateTime = LocalDateTime.now(ZoneOffset.UTC)
```
ElenaSpb, line 56: please use here 
TimeService.now() function

naXa777: @ElenaSpb, I recommend against injecting a service into an entity. It's not the best practice.

Issue #2430, opened 2022-12-29
[BE Investigation] Make one call to S3 to check if list of pictures are exists


ElenaSpb:
[BE Investigation] Make one call to S3 to check if list of pictures are exists

now for each picture/file server make a call to s3 to check if it is exist so in case 20 files  -it would be 20 calls
Is it possible to reduce calls numbers making one cal with files list as parameter.
?

mironovandrei:
@ElenaSpb looks like there is no way to check existence of multiple files in one call(listObjects is not the best way to do that because it'll download all the files matching the conditions). So to optimize this check we can use headObject here as it's recommended by AWS docs(https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_HeadObject_section.html)

lifeart:
@mironovandrei Likely this task is related to https://github.com/Brain-up/brn/issues/2455

mironovandrei:
@lifeart ok, thanks for explanation
anyway, there are no ways to check existence of file but the AWS recommended one(headObject)

lifeart:
@mironovandrei as I understand, we could just get list of all files and work with it https://bobbyhadz.com/blog/aws-cli-list-all-files-in-bucket

mironovandrei:
@lifeart my bad, listObjects is good enough for this task
I'll use it

ElenaSpb:
implementation by start everyday job which update resource pictures, so no need to check it each time before using, it already exists in Resource entity

PR #2489, opened 2023-02-20
[WIP]2430 check file existence in S3 in according to AWS documentation

mironovandrei:
#2430 
Use AWS recommended way to check if file exists in S3 bucket or not

### DIFF
--- a/src/main/kotlin/com/epam/brn/service/cloud/AwsCloudService.kt
+++ b/src/main/kotlin/com/epam/brn/service/cloud/AwsCloudService.kt
@@ -121,6 +121,26 @@ class AwsCloudService(@Autowired private val awsConfig: AwsConfig, @Autowired pr
         }
     }
 
+    override fun findExistingFiles(filePath: String, words: List<String?>, extensions: Set<String>): Map<String, String> {
+        val result = mutableMapOf<String, String>()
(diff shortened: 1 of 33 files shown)


src/main/kotlin/com/epam/brn/service/cloud/AwsCloudService.kt
```diff
@@ -107,13 +108,37 @@ class AwsCloudService(@Autowired private val awsConfig: AwsConfig, @Autowired pr
...
         val fullFileName = createFullFileName(filePath, fileName)
+        // Recommended way to check file existence according to AWS documentation https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_HeadObject_section.html
+        return try {
+            val request = HeadObjectRequest.builder()
+                .bucket(awsConfig.bucketName)
+                .key(fullFileName)
+                .build()
+            s3Client.headObject(request)
+            true
+        } catch (e: NoSuchKeyException) {
```
naXa777, line 119: The documentation mentions `S3Exception`. I don't know if this is the recommendation or just an example. Anyway why did you use `NoSuchKeyException` instead?

ElenaSpb:
Thank you !
we implement another solution by using ResourcePictureUrlUpdateJob to set resource picture

PR #2527, opened 2023-10-19
#2406 Add lastVisit field and set it when user logs in

### LINKED ISSUE #2406: [BE] Add new field LastVisit field in https://brainup.fun/admin/users response
[BE] AddLast visit field in https://brainup.fun/admin/users response that will be about last visit to our site.
not about last done exercises as it is now. current value from Last Visit move to field Last study activity

image

naXa777:
Closes #2406 "Add lastVisit field and set it when user logs in"

Changes description:
* I pulled PR with unfinished work #2444
    * v1 of this task was implemented by @chackomathew
* I exposed new `lastVisit` field in `UserWithAnalyticsResponse` returned by `GET /users?withAnalytics=true` API
* The registration date is used as "Last Visit" date in API response, if last visit date is unknown.
* Code refactoring.
* Integration with admin UI.


src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
```diff
@@ -73,6 +75,13 @@ class UserAccountServiceImpl(
             .orElseThrow { EntityNotFoundException("No user was found for email=$email") }
     }
 
+    override fun markVisitForCurrentUser() {
+        getCurrentUser().let {
+            it.lastVisit = timeService.now()
+            userAccountRepository.save(it)
+        }
```
naXa777, line 82: the integration tests are failing because the new code now makes a query to get current user from DB, and fails, cause test DB doesn't contain test users.

```
No user was found for email=test@test.testcom.epam.brn.exception.EntityNotFoundException: No user was found for email=test@test.test
	at app//com.epam.brn.service.impl.UserAccountServiceImpl.getCurrentUser$lambda-5(UserAccountServiceImpl.kt:75)
	at java.base@15.0.10/java.util.Optional.orElseThrow(Optional.java:401)
	at app//com.epam.brn.service.impl.UserAccountServiceImpl.getCurrentUser(UserAccountServiceImpl.kt:75)
	at app//com.epam.brn.service.impl.UserAccountServiceImpl.markVisitForCurrentUser(UserAccountServiceImpl.kt:79)
```

I consider two alternative solutions:

1. Keep the main code "as is". Prepare user records in test DB: insert test user before test, delete test users after test.
2. Change `fun markVisitForCurrentUser()` in the main code, so instead of SELECT+UPDATE here we only do UPDATE of a single column:
    ```
    val authentication = SecurityContextHolder.getContext().authentication
    val email = authentication.name ?: getNameFromPrincipals(authentication)
    userAccountRepository.updateLastVisit(email, timeService.now())
    ```
    if user account with specified email doesn't exist - don't throw error, ignore.

ElenaSpb:
could you please join our dev chat 
https://t.me/+R-6ThlxgP5QyZmMy

src/main/kotlin/com/epam/brn/service/impl/UserAccountServiceImpl.kt
naXa777: 2

src/main/kotlin/com/epam/brn/auth/filter/FirebaseTokenAuthenticationFilter.kt
```diff
@@ -37,9 +37,16 @@ class FirebaseTokenAuthenticationFilter(
...
     ) {
         verifyToken(request)
+        markVisit()
         filterChain.doFilter(request, response)
     }
 
+    private fun markVisit() {
+        if (SecurityContextHolder.getContext().authentication != null) {
+            userAccountService.markVisitForCurrentUser()
+        }
```
ElenaSpb, line 47: we dont use {} after condition with one code line
https://github.com/Brain-up/brn/wiki/Coding-Standards

src/main/resources/db/migration/V220230119_2406.sql
```diff
@@ -0,0 +1 @@
+alter table user_account add column if not exists last_visit timestamp;
```
ElenaSpb, line 1: you should create new script

from readme:
[Flyway scripts naming](https://github.com/Brain-up/brn#flyway-scripts-naming)
use V2yearmonthday_taskNumber for example V220210804_899.

src/main/kotlin/com/epam/brn/auth/filter/FirebaseTokenAuthenticationFilter.kt
naXa777: > To verify that your code is formatted according to the style guide, go to the inspection settings and enable the "Kotlin | Style issues | File is not formatted according to project settings" inspection.

The wiki page is partially outdated. I don't have "File is not formatted according to project settings" inspection in Settings. The new inspection is: "General -> Incorrect Formatting." And it was disabled by default.

image

naXa777: > we do not use extra brackets {} in single line after condition

* This rule can't be enforced by IntelliJ IDEA. I spent some time on trying to do so, and I feel frustrated now.
* This rule is not enforced by ktlint.
* And I see some code that violates this rule in the codebase. For example, [`verifyFileExtension`](https://github.com/Brain-up/brn/commit/3a7704fe262d0d6ba2749211d286b7063252af6c#diff-7572837f9bace6bc00f1eea7c29932cb19809a3003bb60f1889aa05217b94b9eR78) function.

naXa777: I've removed curly braces

src/main/resources/db/migration/V220230119_2406.sql
naXa777: I've renamed the script

naXa777:
> I think it would be logically to set lastVisit in case null - the registration date. it can not be null

Done.

naXa777:
> "current value from Last Visit move to field lastStudyActivity"
> do you want to implement it too?

yes, but I don't see "Last Visit" field anywhere in code. There was no Last Visit field before, or maybe it was removed recently, IDK. @ElenaSpb, please clarify this requirement.

naXa777:
some checks have failed. and I suspect that they are broken, cause it's not related to the changes in PR #2527.

I kindly ask to merge this PR and fix the failing GitHub actions separately:
* #2531
* #2532

lifeart (approved):

ElenaSpb (approved):

ElenaSpb:
Thank you for contribution to our project!

PR #2541, opened 2023-11-27
fix: only numbers allowed in year of birth

iajaymk:
##2474

I added a keydown method to only accept numeric values for the year of birth input on both the registration and profile pages.


lifeart (approved):

lifeart:
@iajaymk thank you for contribution!

PR #2545, opened 2023-11-29
2544 add accounts for auto tests

ElenaSpb:
##2544

add accounts for auto tests


ElenaSpb:
add accounts for auto tests

PR #2559, opened 2023-12-23
2556 [BE] get autotest contributors from repo https://github.com/Brai…

andrsam:
2556 [BE] get autotest contributors from repo


src/main/kotlin/com/epam/brn/service/ContributorServiceImpl.kt
```diff
@@ -68,6 +68,11 @@ class ContributorServiceImpl(
             name = gitHubUser.name ?: "gitHubNick:${gitHubUser.login}",
             repositoryName = repositoryName,
             company = gitHubUser.company,
+            type = if (repositoryName == "auto-tests-python") {
+                ContributorType.AUTOTESTER
+            } else {
```
ElenaSpb, line 73: we dont use {} n single conditions, please

ElenaSpb (approved):

Issue #2564, opened 2024-01-13
[BUG UI] The “Description” menu item is duplicated by the “Information” item in the navbar 


ElenaSpb:
![image](https://github.com/Brain-up/brn/assets/15213856/70f64214-dcec-4fa5-9381-43dc97bed229)
 
delete Информация from navbar

PR #2567, opened 2024-01-17
bug: issue #2564

iajaymk:
##2564

Removed 'Info' from the more menu in the header.


lifeart:
@iajaymk thank you for contribution!

Issue #2568, opened 2024-01-20
[Bug] There is no validation message that the user must accept the terms of the user agreement in Registration page


Nat754:
Preconditions: If necessary

Reproduction steps:
1. Open the page https://brainup.site/
2. Click the button "Registration"
3. Fill in the fields with the correct data  
4. The accept checkbox is not selected
5. Click the button "Registration"

Expected Result:
There is validation message that the user must accept the terms of the user agreement

Actual result:
There is no validation message that the user must accept the terms of the user agreement

Application:
![image](https://github.com/Brain-up/brn/assets/71374230/48012680-a194-45cb-8f9f-0ab9f57226f6)
Screenshot, skincast, link to Allure report

Environment:
 - OS: [Windows 10 x 64]
 - Browser [Chrome]
 - Version [120.0.6099.225]

Additional context: Add any other context about the problem here.
There is no validation message if user doesn't fill data and click the button Registration

https://github.com/Brain-up/brn/assets/71374230/3ff3f6e7-4f44-43a7-94c6-24b41718ca21

PR #2656, opened 2024-12-11
registration-form: add validation for agreed, add values in template …

MSt1ch:
…for agreed and gender components

##GithubTaskNumber

### DIFF
--- a/frontend/app/components/registration-form/component.ts
+++ b/frontend/app/components/registration-form/component.ts
@@ -32,6 +32,9 @@ export default class RegistrationFormComponent extends LoginFormComponent {
   @tracked gender!: 'MALE' | 'FEMALE';
   @tracked agreed = false;
 
+  @tracked agreedStatusErrorMessage = '';
+  @tracked serverErrorMessage = '';
+
   get warningPasswordsEquality() {
     if (this.repeatPassword === undefined) {
       return false;
@@ -110,7 +113,7 @@ export default class RegistrationFormComponent extends LoginFormComponent {
       ) as FirebaseAuthenticator;
       yield auth.registerUser(user.email, user.password);
     } catch (e) {
-      this.errorMessage = e.message;
+      this.serverErrorMessage = e.message;
       yield this.registrationTask.cancelAll();
       return;
     }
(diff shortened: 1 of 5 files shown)


frontend/app/components/registration-form/template.hbs
```diff
@@ -138,11 +146,16 @@
...
+      {{#if this.agreedStatusErrorMessage}}
+        <p data-test-form-error class="mt-2 text-xs text-red-500">
+          {{this.agreedStatusErrorMessage}}
+        </p>
+      {{/if}}
       <div class="flex mb-4">
         <Ui::Button
           @type="submit"
           class="w-full"
-          disabled={{not this.agreed}}
```
lifeart, line 145: I think we still need this logic

frontend/app/components/registration-form/component.ts
```diff
@@ -32,6 +32,9 @@ export default class RegistrationFormComponent extends LoginFormComponent {
   @tracked gender!: 'MALE' | 'FEMALE';
   @tracked agreed = false;
 
+  @tracked agreedStatusErrorMessage = '';
```
lifeart, line 35: let's rewrite it to getter, because it's completely depends on `agreed` value:
```ts
get agreedStatusErrorMessage() {
   return this.agreed ? '' : this.intl.t('registration_form.agree_terms');
}


```

frontend/app/components/registration-form/component.ts
```diff
@@ -82,6 +85,7 @@ export default class RegistrationFormComponent extends LoginFormComponent {
   }
 
   get registrationInProgress() {
+    console.log('here')
```
lifeart, line 88: ~console.log('here')~

frontend/app/components/registration-form/template.hbs
```diff
@@ -68,6 +69,7 @@
           value="MALE"
           class="w-3 h-3 ml-4"
           id="male"
+          checked={{eq this.gender "MALE"}}
```
lifeart, line 72: good improvement!

frontend/app/components/registration-form/component.ts
MSt1ch: oops)

frontend/app/components/registration-form/component.ts
MSt1ch: I'm not sure here. You can write the issue below
https://github.com/Brain-up/brn/issues/2568

if we use the getter, an error will immediately appear on the form. Although we haven't started filling it out yet. In this case, I show it only after all the validations and if you don't accept the checkbox.

lifeart: Agree, make sense

lifeart (approved):

lifeart:
@MSt1ch thank you for contribution!

PR #2673, opened 2025-03-24
BE Add get all resources end-point

ElenaSpb:
##2672

Add get all resources end-point


ElenaSpb:
Add get all resources end-point

PR #2692, opened 2025-04-04
ci: db backup fixes #2418

### LINKED ISSUE #2418: [DevOps] Make backup of our DB every night
[DevOps] Make backup of our DB every night

save only for last week, others backup should be deleted

marynik:
#2418

- Introduces a new github action on selectel runner
- makes database backup and gzip it
- sends the gzipped file to s3 bucket

### DIFF
--- a/.github/workflows/db-backup.yml
+++ b/.github/workflows/db-backup.yml
@@ -1,0 +1,46 @@
+name: Makes DB backup every week
+
+on:
+  schedule:
+    - cron: "0 4 * * 6"
+  workflow_dispatch:
+
+jobs:
+  db_backup:
+    runs-on: selectel
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Prep docker-compose.yml
+        run: mv docker-compose-run.yml docker-compose.yml
+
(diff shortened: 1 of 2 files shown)


DmitriyStoyanov:
What the process of restoration will be here if it will be required? Possibly need to write some documentation about that?

marynik:
> What the process of restoration will be here if it will be required? Possibly need to write some documentation about that?

I supposed it to be manual. Is it needed to write the docs about the restoration process?

marynik:
@DmitriyStoyanov I considered creating a GitHub Action for restoring, but decided against it — one-click restore could lead to accidental rollbacks and data loss. I think full DB restore should stay manual. The main use case will likely be partial restores, which are hard to automate. So I’d suggest not automating it at all. We can document the process instead. Let me know what you think.

DmitriyStoyanov:
Better to have documentation about the process itself, instead of adding it into github actions.

marynik:
@DmitriyStoyanov so let's merge this PR? And configure the s3 bucket and GitHub Actions secret, so that I could write some docs regarding this

PR #2704, opened 2025-04-14
Issue-2695 Replace kluent by kotest

andresdigital3001:
##GithubTaskNumber

Delete Kluent and replace it by kotest. https://github.com/Brain-up/brn/issues/2695


andresdigital3001:
This was replaced by another because of standard names

PR #2752, opened 2025-09-08
2595 - add active field into series model, add tests for new field

PatrycjaBuczkowskaEPAM:
2595

I added new active field into Series model. I also added tests for new field

### DIFF
--- a/src/main/kotlin/com/epam/brn/model/Series.kt
+++ b/src/main/kotlin/com/epam/brn/model/Series.kt
@@ -13,7 +13,6 @@ import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
 import javax.persistence.OneToMany
-import java.lang.Boolean.TRUE
 
 @Entity
 class Series(
@@ -27,7 +26,7 @@ class Series(
     var level: Int,
     var description: String? = "",
     @Column
-    var active: Boolean = TRUE,
+    var active: Boolean = true,
     @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "exercise_group_id")
     var exerciseGroup: ExerciseGroup,
--- a/src/main/kotlin/com/epam/brn/model/Series.kt
+++ b/src/main/kotlin/com/epam/brn/model/Series.kt
@@ -13,6 +13,7 @@ import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.ManyToOne
 import javax.persistence.OneToMany
+import java.lang.Boolean.TRUE
 
 @Entity
 class Series(
@@ -25,6 +26,8 @@ class Series(
     @Column
     var level: Int,
     var description: String? = "",
+    @Column
+    var active: Boolean = TRUE,
     @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "exercise_group_id")
     var exerciseGroup: ExerciseGroup,
--- a/src/main/kotlin/com/epam/brn/service/SeriesService.kt
+++ b/src/main/kotlin/com/epam/brn/service/SeriesService.kt
@@ -16,7 +16,10 @@ class SeriesService(
     @Cacheable("series")
     fun findSeriesForGroup(groupId: Long): List<SeriesDto> {
         log.debug("try to find series for groupId=$groupId")
-        val series = seriesRepository.findByExerciseGroupLike(groupId)
+        val series =
+            seriesRepository
+                .findByExerciseGroupLike(groupId)
+                .filter { it.active }
         return series.map { seriesEntry -> seriesEntry.toDto() }
     }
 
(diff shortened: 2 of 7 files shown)


src/main/kotlin/com/epam/brn/model/Series.kt
```diff
@@ -25,6 +26,8 @@ class Series(
     @Column
     var level: Int,
     var description: String? = "",
+    @Column
+    var active: Boolean = TRUE,
```
umedvedev, line 30: since new column was added, there should be a new Flyway migration added as well

PatrycjaBuczkowskaEPAM: Removed update statement from script, cause from version 11 of PostgreSQL ALTER TABLE statement with DEFAULT VALUE for new column with add this column with provided default value to all exsisting records

src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -16,7 +16,10 @@ class SeriesService(
     @Cacheable("series")
     fun findSeriesForGroup(groupId: Long): List<SeriesDto> {
         log.debug("try to find series for groupId=$groupId")
-        val series = seriesRepository.findByExerciseGroupLike(groupId)
+        val series =
+            seriesRepository
```
ElenaSpb, line 20: could you please make filter in sql

PatrycjaBuczkowskaEPAM: Sure, I can add it

src/main/kotlin/com/epam/brn/model/Series.kt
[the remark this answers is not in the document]
PatrycjaBuczkowskaEPAM: I changed to lower case

src/main/kotlin/com/epam/brn/service/SeriesService.kt
ElenaSpb: yes

PatrycjaBuczkowskaEPAM: @ElenaSpb I didn't add filtering by active to all methods and added autogenerated methods like `findDistinctByIdAndActiveTrue`, please check if that is what you wanted me to add

src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -15,18 +15,20 @@ class SeriesService(
...
 
     @Cacheable("seriesDto")
     fun findSeriesDtoForId(seriesId: Long): SeriesDto {
-        log.debug("try to find series for seriesId=$seriesId")
+        log.debug("try to find active series for seriesId=$seriesId")
         val series =
             seriesRepository
-                .findById(seriesId)
-                .orElseThrow { EntityNotFoundException("no series was found for id=$seriesId") }
+                .findDistinctByIdAndActiveTrue(seriesId)
```
ElenaSpb, line 30: there findById would be better

PatrycjaBuczkowskaEPAM: I removed distinct from name of the method, cause we search by PK, so we don't need it in here.

src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -15,18 +15,20 @@ class SeriesService(
...
     fun findSeriesDtoForId(seriesId: Long): SeriesDto {
-        log.debug("try to find series for seriesId=$seriesId")
+        log.debug("try to find active series for seriesId=$seriesId")
         val series =
             seriesRepository
-                .findById(seriesId)
-                .orElseThrow { EntityNotFoundException("no series was found for id=$seriesId") }
+                .findByIdAndActiveTrue(seriesId)
+                ?: throw EntityNotFoundException("no active series was found for id=$seriesId")
         return series.toDto()
```
umedvedev, line 32: could it be expressed like:

```kotlin
 return seriesRepository
    .findByIdAndActiveTrue(seriesId)
    ?.toDto()
    ?: throw EntityNotFoundException("No active series was found for id=$seriesId")
```

src/main/kotlin/com/epam/brn/service/SeriesService.kt
```diff
@@ -15,18 +15,20 @@ class SeriesService(
 
     @Cacheable("series")
     fun findSeriesForGroup(groupId: Long): List<SeriesDto> {
-        log.debug("try to find series for groupId=$groupId")
-        val series = seriesRepository.findByExerciseGroupLike(groupId)
+        log.debug("try to find active series for groupId=$groupId")
+        val series =
+            seriesRepository
+                .findDistinctByExerciseGroupIdAndActiveTrue(groupId)
         return series.map { seriesEntry -> seriesEntry.toDto() }
```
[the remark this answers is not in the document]
umedvedev: or even rather

```kotlin
return  seriesRepository.findDistinctByExerciseGroupIdAndActiveTrue(groupId).map { it.toDto() } 
```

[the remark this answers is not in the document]
PatrycjaBuczkowskaEPAM: I changed that and it works. Thanks :)

umedvedev (approved):

ElenaSpb (approved):

